PR #71, opened 2022-02-04
Colorize screen output by log level if available

gulducat:
Nice quick lil pleasantry, really helps WARNs and ERRORs pop.

```go
	// AutoColor checks if the io.Writer is a tty,
	// and if so enables coloring.
	AutoColor
```

https://github.com/hashicorp/go-hclog/blob/e77830785dbe9f30d4d968c68bcb8fe934b9af6e/logger.go#L84-L86


mkcp (approved):
This is nice to have and we should also ensure it can be disabled via CLI flag if for some reason that's necessary.

PR #72, opened 2022-02-09
add semver flag

mkcp:
Probably should've had this on release last year but better late than never!

- Add `-version` to our flags. it prints the name of the binary and the prefixed semver.
- `SemVer` is tracked in a constant in main.go

Behavior of branch:
Screen Shot 2022-02-09 at 10 52 06 AM

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,7 +13,7 @@ import (
 	"github.com/hashicorp/hcdiag/agent"
 )
 
-const SemVer string = "0.1.4"
+const SemVer string = "0.1.3"
 
 func main() {
 	os.Exit(realMain())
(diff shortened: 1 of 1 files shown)


mkcp:
accidentally set the semver to 0.1.4 in the initial, we're actually on patch version 0.1.3

gulducat (approved):
No blockies, definitely nice to have!

Some food for thought: Other HashiCorp products also work without `-`
like `consul version`
Do we want to be consistent with them like that?

mkcp:
That command pattern comes from Consul and other products having "subsystems" in their CLI. We don't branch into multiple sets of commands like they do, so it doesn't make sense to me to add it just for version. If we do add CLI subsystems to help manage complexity, then we should add one for version!

PR #74, opened 2022-02-09
add goclangci lint

mkcp:
I wonder if this will run any actions that are added in its own PR 🤔 

edit:

This adds the golangci-lint action, which has various opinions, config, and defaults that we can use. It gives us line by line integration for diffs, which is nicer than us spinning it ourselves! The first commit fixes every issue raised by running `golangci-lint run` locally, then the remaining PRs modify the actions to make this fit.

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -4,7 +4,6 @@ on:
     tags:
       - v*
     branches:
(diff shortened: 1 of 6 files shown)


mkcp:
The answer is yes!
Screen Shot 2022-02-09 at 12 43 58 PM

PR #76, opened 2022-02-10
Add journald seeker for consul, nomad, vault

gulducat:
A common setup is to use systemd units to run our software as linux services (indeed our own apt packages include such systemd config).  As a result, a common location for log output is journald (via program stdout/stderr).

This attempts to include logs from journald automatically, by "simply" delegating commands to the user's shell.

For example with `hcdiag -vault`:
* detect the presence of a systemd unit named {program} via `systemctl is-enabled vault`
* check whether the user is able to read from journald with a `journalctl` pre-check
  * because journalctl will happily exit `0` "success" even if the user is not actually able to read the logs
* return a Sheller seeker to run ~ `journalctl -u vault > {bundleDir}/journald-vault.log` with the other Vault seekers.

Two hard-coded limitations at present:
* The unit name can not be customized, only "consul", "nomad", or "vault" are supported.
* The timeframe of logs is also static, from "3 days ago"

I started writing tests, but since this implementation just delegates to `systemctl` and `journalctl` commands, they would need mock'd away in some manner, and I haven't quite settled on the approach there.  Happy to chat through it if including tests here is a blocker.

### DIFF
--- a/product/consul.go
+++ b/product/consul.go
@@ -44,6 +44,10 @@ func ConsulSeekers(tmpDir string, from, to time.Time) []*s.Seeker {
 		logCopier := s.NewCopier(logPath, dest, from, to)
 		seekers = append([]*s.Seeker{logCopier}, seekers...)
 	}
+	// get logs from journald if available
+	if journald := s.JournaldGetter("consul", tmpDir); journald != nil {
+		seekers = append(seekers, journald)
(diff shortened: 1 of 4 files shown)


mkcp (approved):
We've reviewed this code synced 👍

PR #77, opened 2022-02-24
Add since flag

mkcp:
Review guide:

- Readme updated
- substantial changes to how time is set in `main.go` including a bugfix for the time.Unix(x,y) which was setting the wrong relative time.
- everything in the agent->product->seeker flow should be simple renames.
- add since support to journald.


gulducat (approved):
phew, we made it!  :shipit: 

since when!

PR #78, opened 2022-02-25
add windows builds back in

mkcp:
Early on we maintained windows builds, but we dropped them due to low demand. Demand is back baby, this adds them back in!


gulducat (approved):
windows cd

PR #82, opened 2022-03-05
Refactor host seeker

davemay99:
This PR refactors the host seekers to:
* simplify control logic
* remove extra code
* improve logging

During the process I discovered that we were missing swap memory as well as process ID / parent ID information.  Those were both addressed here as well but could be split out if necessary.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.17
 
 require (
 	github.com/hashicorp/go-hclog v0.16.2
+	github.com/hashicorp/go-multierror v1.1.1
 	github.com/hashicorp/hcl/v2 v2.10.1
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/go-ps v1.0.0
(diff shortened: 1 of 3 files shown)


mkcp:
@davemay99 We appreciate you taking the time to add error handling to the host seeker. However, we cannot accept this PR as written because it differs quite a bit from the patterns we use in other seekers.

The first commit is a reasonable solution that we can accept. The commits after introduce an abstraction around logs wrapping and relies on shared mutable state on the seeker for storing errors and the result. While this line count is reduced, this adds novel complexity that doesn't fit the design goals we have for seekers.

The host seeker is a bit of an odd case already because it was introduced before we finished the seeker abstraction. All of its operations are glommed onto one special-case seeker as an intermediate step. In an upcoming milestone, we want to improve logging and seeker error handling throughout hcdiag - keeping the host seeker as close as possible to the other seekers makes that effort simpler.

As a path forward, I'm closing this PR and am applying the first commit from the branch onto main. This will let us move onto the proposed additions to the host seeker and changes to Nomad's seekers.

mkcp:
@davemay99 We appreciate you taking the time to add error handling to the host seeker. However, we cannot accept this PR as written because it differs quite a bit from the patterns we use in other seekers.

The first commit is a reasonable solution that we can accept. The commits after introduce an abstraction around logs wrapping and relies on shared mutable state on the seeker for storing errors and the result. While this line count is reduced, this adds novel complexity that doesn't fit the design goals we have for seekers.

The host seeker is a bit of an odd case already because it was introduced before we finished the seeker abstraction. All of its operations are glommed onto one special-case seeker as an intermediate step. In an upcoming milestone, we want to improve logging and seeker error handling throughout hcdiag - keeping the host seeker as close as possible to the other seekers makes that effort simpler.

As a path forward, I'm closing this PR and am applying the first commit from the branch onto main. This will let us move onto the proposed additions to the host seeker and changes to Nomad's seekers.

PR #83, opened 2022-03-23
Nomad seeker updates

mkcp:
PR from pairing with Dave.

Adds multiple sample windows with duration/interval in the support bundle, and changes the api calls to perform potentially stale cluster state reads from the node being read from.


gulducat (approved):
yay!

PR #85, opened 2022-03-30
Necessary bits to onboard project to the Releases API

claire-labry:
These are the necessary components that `hcdiag` needs to onboard to Releases API.


.github/workflows/release.yml
```diff
@@ -26,38 +26,44 @@ jobs:
         id: codesign
         env:
           VERSION: v0
-
-      - uses: hashicorp/setup-hc-releases@v1
+      - name: Setup hc-releases
+        uses: hashicorp/actions-setup-hc-releases@v2
```
jeanneryan, line 30: this should be `setup-hc-releases@v2`, we can drop `actions-`

claire-labry: thank you for doing this! ❤️

jeanneryan (approved):

nwchandler (approved):
LGTM!

PR #93, opened 2022-04-07
Migrate osinfo out of hostseeker and into seeker/host

mkcp:
Let's hold off on merging this until we can discuss our release process more. Do new seekers belong in minor versions instead of patches?


nwchandler (approved):
Looks good to me!

PR #95, opened 2022-04-14
Add error types for copier operations AND add copier test

mkcp:
2 distinct threads here from the same investigation:

Thread 1: Here we create error types for the operations that can fail in Copier.Run(), adding relevant fields like the path and files we attempted to copy.

Thread 2: Also added in a test for NewCopier. I briefly looked into testing fs operations with testing/fstest, but the file operations in `util` are a better place focus unit testing efforts (https://hashicorp.atlassian.net/browse/CORI-51). In the future a cross-platform unit test for Copier.Run() would be valuable.

Example seeker pkg error that wraps the error from `util`:
Screen Shot 2022-04-14 at 11 32 11 AM
 
This shows a bit of inconsistency in our logging between the agent and seekers. We ought to find good reference examples w/ hclog and see how we can match or improve on it!

### DIFF
--- a/seeker/copier.go
+++ b/seeker/copier.go
@@ -1,6 +1,7 @@
 package seeker
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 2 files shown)


gulducat (approved):
nice, extra props for the `Unwrap()`s ^_^

unwrap to find what's hidden inside

PR #98, opened 2022-04-15
Separate host seekers, via new GoFuncSeeker that runs go functions

gulducat:
The idea here is to have a general-purpose seeker that runs any Go function with a particular signature.

It lines up with Commander/Sheller whose jobs are to go run arbitrary CLI commands, Copier which copies arbitrary files, and HTTPer which makes arbitrary http calls.

Some of this reads a bit verbose due to the necessity of specifying the function signature in the inline functions to cause type coercion to `interface{}`, but I think it's a pretty slim amount of boilerplate overall.

I have left one intentionally-bad seeker in here to illustrate a failure mode.  Can you spot it? 😋 

Example output:

```
2022-04-15T10:28:29.502-0500 [INFO]  hcdiag: Checking product availability
2022-04-15T10:28:29.502-0500 [INFO]  hcdiag: Gathering diagnostics
2022-04-15T10:28:29.502-0500 [INFO]  hcdiag: Running seekers for: product=host
2022-04-15T10:28:29.502-0500 [INFO]  hcdiag: running: seeker="uname -v"
2022-04-15T10:28:29.508-0500 [INFO]  hcdiag: running: seeker=host
2022-04-15T10:28:29.561-0500 [INFO]  hcdiag: running: seeker=disks
2022-04-15T10:28:29.562-0500 [INFO]  hcdiag: running: seeker=memory
2022-04-15T10:28:29.562-0500 [INFO]  hcdiag: running: seeker=network
2022-04-15T10:28:29.562-0500 [INFO]  hcdiag: running: seeker=processes
2022-04-15T10:28:29.564-0500 [INFO]  hcdiag: running: seeker=bad-and-not-good
2022-04-15T10:28:29.564-0500 [WARN]  hcdiag: result: seeker=bad-and-not-good result=%!s(<nil>) error="no good at all"
2022-04-15T10:28:29.564-0500 [INFO]  hcdiag: Product done: product=host statuses="map[fail:1 success:6]"
2022-04-15T10:28:29.566-0500 [INFO]  hcdiag: Created Results.json file: dest=hcdiag-2022-04-15T152829Z/Results.json
2022-04-15T10:28:29.566-0500 [INFO]  hcdiag: Created Manifest.json file: dest=hcdiag-2022-04-15T152829Z/Manifest.json
2022-04-15T10:28:29.570-0500 [INFO]  hcdiag: Compressed and archived output file: dest=hcdiag-2022-04-15T152829Z.tar.gz
```

Results.json snippet:

```json
{
    "host": {
        "bad-and-not-good": {
            "runner": {
                "Name": "github.com/hashicorp/hcdiag/product.NewHost.func5"
            },
            "result": null,
            "error": "no good at all",
            "status": "fail"
        },
        "disks": {
            "runner": {
                "Name": "github.com/hashicorp/hcdiag/product.NewHost.func2"
            },
            "result": [
                {
                    "device": "/dev/disk1s5s1",
                    "mountpoint": "/",
                    "fstype": "apfs",
...
        "processes": {
            "runner": {
                "Name": "github.com/hashicorp/hcdiag/product.GetProcesses"
            },
            "result": [
                "hcdiag",
...
```


gulducat:
I wonder if we have some bad math somewhere hah, test failure:

```
2022-04-15T15:42:39.512Z [WARN]  result: seeker=bad-and-not-good result=%!s(<nil>) error="no good at all"
2022-04-15T15:42:39.512Z [INFO]  Product done: product=host statuses="map[fail:1 success:6]"
    agent_test.go:153: 
        	Error Trace:	agent_test.go:153
        	Error:      	Received unexpected error:
        	            	more than 50% of Seekers failed
        	Test:       	TestRunProducts
--- FAIL: TestRunProducts (0.01s)
```

mkcp:
The impl. is clean, but it doesn't get us closer to longer term goals for seekers. While it trims some boilerplate, it also means that functions must take 0 params and they cannot interpret errors. While this means they match other seekers in the results, internally they remain a special case. We'd be better served by making host seekers consistent with other seekers, boilerplate and all, so they're easier to grow, change, and include in configuration in the future.

mkcp:
> I wonder if we have some bad math somewhere hah, test failure:
> 
> ```
> 2022-04-15T15:42:39.512Z [WARN]  result: seeker=bad-and-not-good result=%!s(<nil>) error="no good at all"
> 2022-04-15T15:42:39.512Z [INFO]  Product done: product=host statuses="map[fail:1 success:6]"
>     agent_test.go:153: 
>         	Error Trace:	agent_test.go:153
>         	Error:      	Received unexpected error:
>         	            	more than 50% of Seekers failed
>         	Test:       	TestRunProducts
> --- FAIL: TestRunProducts (0.01s)
> ```

The math looks ok - maybe this is concurrency bug? I didn't realize until now that we were incrementing a non-atomic int as `a.NumErrors++` in a.runSet() for each product, so the outcome of that is indeterminate. Rather than making a.NumErrors threadsafe we're better off making the critical path simpler by removing it.

gulducat:
thanks for feedback!  as a preface, the strong feelings that I do seem to have are not actually that important, so I will defer to your judgement as you seem to have a firmer vision of what the boilerplate in #96 may enable in the future.  that said, here are my thoughts:

> functions must take 0 params and they cannot interpret errors

this isn't strictly true, I just hadn't implemented those things pre-emptively.  [2b91656](https://github.com/hashicorp/hcdiag/pull/98/commits/2b91656c53ae41aa95a8f90513cafb4c63a9397c) splits all of them out into the kind of boilerplate one could use if desired to enable error-classification and status per GoFunc, and a param on DiskPartitions by returning a closure function that itself is niladic.

> making host seekers consistent with other seekers

the implication is that this is less consistent, which isn't quite true either -- our two implementations for host-info-seeking are just differently-inconsistent with existing seekers.  in your case it's the creation of N new seekers that are not general-purpose so diverge with tradition in that sense, and in my case, it's a new general-purpose function-runner to go with the other general-purpose seekers, but with quite a different mechanism running a new category of functions.

I definitely have a strong preference for generalizing things for various reasons, and as we've chatted about in the past, a higher tolerance for sorta-meta-programmy sometimes overly "clever" kinds of stuff.  I'm curious what others may think, but I think either approach will do just fine, so I won't kick and scream toooo much 😜

PR #100, opened 2022-04-20
add ManifestSeekers with ID, error, and Status

mkcp:
Here we add a `ManifestSeeker` type containing seeker IDs, errors, and Statuses. Then we add `Agent.RecordManifest()` which stores a `ManifestSeeker` for each seeker run in each product. This allows us to inspect seeker statuses and errors at a glance. If the error alone is insufficient for understanding what happened in the seeker, then that gives us valuable info about another error case to handle.

In the future we can use this structure to save info about seeker duration and any other fields that help us understand seekers while avoiding any sensitive data.

Here's what the data looks like for `host` and `consul`
Screen Shot 2022-04-20 at 1 21 50 PM

entire manifest 
```
{
    "started_at": "2022-04-20T13:20:55.707717-07:00",
    "ended_at": "2022-04-20T13:21:08.125642-07:00",
    "duration": "12.417858688 seconds",
    "num_errors": 1,
    "num_seekers": 10,
    "configuration": {
        "host_config": null,
        "products_config": null,
        "operating_system": "auto",
        "serial": false,
        "dry_run": false,
        "consul_enabled": true,
        "nomad_enabled": false,
        "terraform_ent_enabled": false,
        "vault_enabled": false,
        "since": "2022-04-17T13:20:55.707503-07:00",
        "until": "0001-01-01T00:00:00Z",
        "includes": null,
        "destination": "."
    },
    "seekers": {
        "consul": [
            {
                "seeker": "consul version",
                "error": "",
                "status": "success"
            },
            {
                "seeker": "consul debug -output=hcdiag-2022-04-20T202055Z/ConsulDebug -duration=10s -interval=5s",
                "error": "",
                "status": "success"
            },
            {
                "seeker": "GET /v1/agent/self",
                "error": "",
                "status": "success"
            },
            {
                "seeker": "GET /v1/agent/metrics",
                "error": "",
                "status": "success"
            },
            {
                "seeker": "GET /v1/catalog/datacenters",
                "error": "",
                "status": "success"
            },
            {
                "seeker": "GET /v1/catalog/services",
                "error": "",
                "status": "success"
            },
            {
                "seeker": "GET /v1/namespace",
                "error": "404 Not Found",
                "status": "unknown"
            },
            {
                "seeker": "GET /v1/status/leader",
                "error": "",
                "status": "success"
            },
            {
                "seeker": "GET /v1/status/peers",
                "error": "",
                "status": "success"
            }
        ],
        "host": [
            {
                "seeker": "stats",
                "error": "",
                "status": "success"
            }
        ]
    }
}
```

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -143,7 +143,6 @@ func TestRunProducts(t *testing.T) {
 	pCfg := product.Config{OS: "auto"}
 	p := make(map[string]*product.Product)
 	p["host"] = product.NewHost(pCfg)
-	// REVIEW(mkcp): Should we adapt this to call NewAgent? This seems like a symptom of poor boundaries between Agent and Products
 	a := Agent{
 		l:        hclog.Default(),
 		products: p,
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -143,6 +143,7 @@ func TestRunProducts(t *testing.T) {
 	pCfg := product.Config{OS: "auto"}
 	p := make(map[string]*product.Product)
 	p["host"] = product.NewHost(pCfg)
+	// REVIEW(mkcp): Should we adapt this to call NewAgent? This seems like a symptom of poor boundaries between Agent and Products
 	a := Agent{
 		l:        hclog.Default(),
 		products: p,
@@ -153,6 +154,23 @@ func TestRunProducts(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Len(t, a.products, 1, "has one product")
 	assert.NotNil(t, a.products["host"], "product is under \"host\" key")
+}
+
+func TestAgent_RecordManifest(t *testing.T) {
+	t.Run("adds to MetadataSeekers when seekers exist", func(t *testing.T) {
+		// Setup
+		testProduct := "host"
+		a := NewAgent(Config{}, hclog.Default())
+		pCfg := product.Config{OS: "auto"}
+		p := make(map[string]*product.Product)
+		p[testProduct] = product.NewHost(pCfg)
+		a.products = p
+		assert.NotEmptyf(t, a.products[testProduct].Seekers, "test setup failure, no seekers available")
+
+		// Record and check
+		a.RecordManifest()
+		assert.NotEmptyf(t, a.ManifestSeekers, "no seekers metadata added to manifest")
+	})
 }
 
 func TestWriteOutput(t *testing.T) {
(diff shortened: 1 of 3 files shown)


mkcp:
The test case here is a bit reachy, with the agent checking into products for their seekers. I decided to keep it consistent with other agent methods for now, but we should expect to refactor this when we clean up agent boundaries. (Maybe agent calls each product to report its own seekers?)

agent/agent_test.go
```diff
@@ -143,6 +143,7 @@ func TestRunProducts(t *testing.T) {
 	pCfg := product.Config{OS: "auto"}
 	p := make(map[string]*product.Product)
 	p["host"] = product.NewHost(pCfg)
+	// REVIEW(mkcp): Should we adapt this to call NewAgent? This seems like a symptom of poor boundaries between Agent and Products
```
mkcp, line 146: This would be in scope for agent cleanup -- the `agent_test.go` definitely has some signs of "greenfield development" and could be improved.

nwchandler: I agree; it seems like it could be a call to NewAgent or else a test helper that could create an agent with a host product, since we use that in this test and the new one below.

agent/agent_test.go
```diff
@@ -155,6 +156,23 @@ func TestRunProducts(t *testing.T) {
 	assert.NotNil(t, a.products["host"], "product is under \"host\" key")
 }
 
+func TestAgent_RecordManifest(t *testing.T) {
+	t.Run("adds to MetadataSeekers when seekers exist", func(t *testing.T) {
```
nwchandler, line 160: Not sure if it's worthwhile for now or something for the backlog, but would this be a candidate for a table test where we could verify any failure cases as well? (We'd probably have to do some mocking in that case too.)

nwchandler (approved):
I left a couple of comments for consideration, but nothing that would cause me to block merging.

agent/agent_test.go
mkcp: That's a good q - I ran through an implementation of this before pushing the branch up and found that the agent setup code was too elaborate. It worked, but implementing more cases with the pattern highlighted the agent/product/seeker boundary issues more than they tested semantics. We'll fix this problem in a better way down the line.

agent/agent_test.go
mkcp: Added an issue for it [CORI-271](https://hashicorp.atlassian.net/browse/CORI-271)

PR #101, opened 2022-04-20
update to v0.3.0

mkcp:
We missed on main after v0.2.0 (but it is on the release branch) so I figure we skip v0.2.0 on main now and rolling forward to v0.3.0 for ongoing work.


nwchandler (approved):
LGTM!

PR #103, opened 2022-04-21
Allow TLS Configuration on API Calls

nwchandler:
This merge allows Consul, Vault, and Nomad API calls to use TLS, with configuration settings provided using standard environment variables for each product. It builds on the work previously merged from PR #99.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -401,7 +401,10 @@ func (a *Agent) Setup() (map[string]*product.Product, error) {
 	}
 	p := make(map[string]*product.Product)
 	if a.Config.Consul {
-		newConsul := product.NewConsul(cfg)
+		newConsul, err := product.NewConsul(cfg)
+		if err != nil {
+			return nil, err
+		}
 		if consul != nil {
 			customSeekers, err := customSeekers(consul, a.tmpDir)
 			if err != nil {
@@ -415,7 +418,10 @@ func (a *Agent) Setup() (map[string]*product.Product, error) {
 
 	}
 	if a.Config.Nomad {
-		newNomad := product.NewNomad(cfg)
+		newNomad, err := product.NewNomad(cfg)
+		if err != nil {
+			return nil, err
+		}
 		if nomad != nil {
 			customSeekers, err := customSeekers(nomad, a.tmpDir)
 			if err != nil {
@@ -548,7 +554,7 @@ func customSeekers(cfg *ProductConfig, tmpDir string) ([]*seeker.Seeker, error) 
 	case product.Consul:
 		c, err = client.NewConsulAPI()
 	case product.Nomad:
-		c = client.NewNomadAPI()
+		c, err = client.NewNomadAPI()
 	case product.TFE:
 		c = client.NewTFEAPI()
 	case product.Vault:
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -546,7 +546,7 @@ func customSeekers(cfg *ProductConfig, tmpDir string) ([]*seeker.Seeker, error) 
 	var err error
 	switch cfg.Name {
 	case product.Consul:
-		c = client.NewConsulAPI()
+		c, err = client.NewConsulAPI()
 	case product.Nomad:
 		c = client.NewNomadAPI()
 	case product.TFE:
--- a/client/apiclient.go
+++ b/client/apiclient.go
@@ -51,15 +51,19 @@ type TLSConfig struct {
 
 // NewAPIClient gets an API client for a product.
 func NewAPIClient(product, baseURL string, headers map[string]string, t TLSConfig) (*APIClient, error) {
-	tlsClientConfig, err := createTLSClientConfig(t)
-	if err != nil {
-		return nil, err
+	transport := http.DefaultTransport.(*http.Transport)
+
+	// If TLSConfig is set, then we need to configure the TLSClientConfig on the transport.
+	if !reflect.DeepEqual(t, TLSConfig{}) {
+		tlsClientConfig, err := createTLSClientConfig(t)
+		if err != nil {
+			return nil, err
+		}
+		transport.TLSClientConfig = tlsClientConfig
 	}
 
 	client := &http.Client{
-		Transport: &http.Transport{
-			TLSClientConfig: tlsClientConfig,
-		},
+		Transport: transport,
 	}
 
 	return &APIClient{
--- a/client/apiclient.go
+++ b/client/apiclient.go
@@ -57,7 +57,8 @@ func (t TLSConfig) IsDefault() bool {
 		len(t.CACertBytes) == 0 &&
 		t.ClientCert == "" &&
 		t.TLSServerName == "" &&
-		t.Insecure == false
+		// Check whether Insecure is false
+		!t.Insecure
 }
 
 // NewAPIClient gets an API client for a product.
--- a/client/apiclient.go
+++ b/client/apiclient.go
@@ -49,36 +49,83 @@ type TLSConfig struct {
 	Insecure bool
 }
 
-// IsDefault checks whether a TLSConfig structure has had any fields changed from their defaults.
-// If any field is non-default, it returns false; otherwise, it returns true.
-func (t TLSConfig) IsDefault() bool {
-	return t.CACert == "" &&
-		t.CAPath == "" &&
-		t.ClientKey == "" &&
-		len(t.CACertBytes) == 0 &&
-		t.ClientCert == "" &&
-		t.TLSServerName == "" &&
-		// Check whether Insecure is false
-		!t.Insecure
-}
-
 // NewAPIClient gets an API client for a product.
-func NewAPIClient(product, baseURL string, headers map[string]string) *APIClient {
+func NewAPIClient(product, baseURL string, headers map[string]string, t TLSConfig) (*APIClient, error) {
+	tlsClientConfig, err := createTLSClientConfig(t)
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsClientConfig,
+		},
+	}
+
 	return &APIClient{
 		Product: product,
 		BaseURL: baseURL,
 		headers: headers,
-		http:    http.DefaultClient,
-	}
+		http:    client,
+	}, nil
 }
 
-func NewAPIClientFromHTTP(product, baseURL string, headers map[string]string, httpClient HTTPClient) *APIClient {
-	return &APIClient{
-		Product: product,
-		BaseURL: baseURL,
-		headers: headers,
-		http:    httpClient,
+func createTLSClientConfig(t TLSConfig) (*tls.Config, error) {
+	tlsClientConfig := &tls.Config{}
+
+	// If the input TLSConfig object is default, we do not need to update tlsClientConfig.
+	if reflect.DeepEqual(t, TLSConfig{}) {
+		return tlsClientConfig, nil
 	}
+
+	if t.CACert != "" || len(t.CACertBytes) != 0 || t.CAPath != "" {
+		rootConfig := &rootcerts.Config{
+			CAFile:        t.CACert,
+			CACertificate: t.CACertBytes,
+			CAPath:        t.CAPath,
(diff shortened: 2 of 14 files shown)


client/apiclient.go
```diff
@@ -46,6 +48,18 @@ type TLSConfig struct {
 	Insecure bool
 }
 
+// IsDefault checks whether a TLSConfig structure has had any fields changed from their defaults.
+// If any field is non-default, it returns false; otherwise, it returns true.
+func (t TLSConfig) IsDefault() bool {
```
mkcp, line 53: As long as the struct doesn't need default state set in a `New{$StructName}` function  go you can check if `t` is equal to the struct type with no values set: e.g. `if t == TLSConfig{} {...}`.

For checking equality with nested collections (slice, map) and struct, `reflect.DeepEqual()` will do the job so long as all the values are truly equal rather than semantically. 

Here's an example, I believe we can use reflect.DeepEqual() in this case. https://go.dev/play/p/U9ycWaVh1mO

I've also been reading that "github.com/google/go-cmp/cmp"'s `cmp.Equal()` is useful for checking semantic equality (e.g. two different slices with the same values should be equal), but I have not used it yet!

agent/agent.go
```diff
@@ -546,9 +552,9 @@ func customSeekers(cfg *ProductConfig, tmpDir string) ([]*seeker.Seeker, error)
 	var err error
 	switch cfg.Name {
 	case product.Consul:
-		c = client.NewConsulAPI()
+		c, err = client.NewConsulAPI()
 	case product.Nomad:
-		c = client.NewNomadAPI()
+		c, err = client.NewNomadAPI()
```
mkcp, line 557: Ensure err is handled

agent/agent.go
```diff
@@ -546,9 +552,9 @@ func customSeekers(cfg *ProductConfig, tmpDir string) ([]*seeker.Seeker, error)
 	var err error
 	switch cfg.Name {
 	case product.Consul:
-		c = client.NewConsulAPI()
+		c, err = client.NewConsulAPI()
```
mkcp, line 555: Ensure err is handled

nwchandler: I think it should be handled on [line 563](https://github.com/hashicorp/hcdiag/pull/103/files/896abcf9658fda690a0fd9dfeb6e0887f4aaca29#diff-62b6ad581fe3a3059ae8c85ef0f31dde4092bfdecfa7d6857c470bcacaa8cc8bR563).

agent/agent.go
nwchandler: I think it should be handled on [line 563](https://github.com/hashicorp/hcdiag/pull/103/files/896abcf9658fda690a0fd9dfeb6e0887f4aaca29#diff-62b6ad581fe3a3059ae8c85ef0f31dde4092bfdecfa7d6857c470bcacaa8cc8bR563).

client/apiclient.go
nwchandler: Thanks! I tried the `if t == TLSConfig{}` approach, but realized there is a `[]byte` in the struct, which prevents direct comparison. I hadn't considered `reflect.DeepEqual()` - I'll use that instead!

go-cmp sounds useful! For the equality check in this particular case, I think I'll stick with reflection, but it sounds like it could be handy to keep in mind for the future!

client/vault.go
```diff
@@ -48,7 +49,19 @@ func NewVaultAPI() (*APIClient, error) {
...
 
-	return NewAPIClient("vault", addr, headers), nil
+	tlsConfig, err := NewVaultTLSConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	apiClient := NewAPIClient("vault", addr, headers)
+
+	err = configureHttpClientTLS(apiClient.http.(*http.Client), tlsConfig)
```
mkcp, line 59: Agreed on your plan to pass the `TLSConfig` into `NewAPIClient()` on instantiation rather than mutating it by reference after. This would be a blocking review item if not!

product/consul.go
```diff
@@ -15,15 +15,22 @@ const (
...
+	return &Product{
+		Seekers: seekers,
+	}, nil
 }
 
 // ConsulSeekers seek information about Consul.
-func ConsulSeekers(tmpDir string, since, until time.Time) []*s.Seeker {
-	api := client.NewConsulAPI()
+func ConsulSeekers(tmpDir string, since, until time.Time) ([]*s.Seeker, error) {
+	api, err := client.NewConsulAPI()
```
mkcp, line 30: We ought to be passing the `APIClient` into the `<PRODUCT>Seekers` function rather than building it in-place. It will set us up for the future when we migrate ENV var reads to main.go and product-specific TLS fields can be set on `product.Config` in the Agent and passed into products. (💭 And perhaps overridden via .hcl?)

Do you feel that passing the `APIClient` into the `<PRODUCT>Seekers()` fns would be in scope for this set of changes?

client/consul.go
```diff
@@ -34,19 +34,39 @@ func NewConsulAPI() (*APIClient, error) {
...
+	httpClient, err := NewHTTPClientWithTLS(tlsConfig)
+	if err != nil {
+		return nil, err
+	}
 
-	err = configureHttpClientTLS(apiClient.http.(*http.Client), tlsConfig)
+	return httpClient, nil
+}
+
+func NewHTTPClientWithTLS(t *TLSConfig) (*http.Client, error) {
```
nwchandler, line 61: This isn't the file where this function would ultimately live, but I stuck it here to start with and forgot to move it before pushing up a work in progress commit.

agent/agent.go
mkcp: I believe that if the product config isn't set then branch would get skipped. This is a good example of a gotcha to watch out for in go errors, where assigning to the same `err` var multiple times can cause errors to get overwritten and lost without being explicitly handled.

product/consul.go
nwchandler: I completely agree! It feels like it could be slightly outside of scope, and since the PR is already kind of big, I think I'll write up a backlog item for this.

agent/agent.go
nwchandler: I'm afraid I'm missing something. 😟  Since go only evaluates one case (and doesn't need the explicit `break` from C), I thought only one case would be hit, so err would only be assigned one value. I could see issues potentially starting around line 567 if none of the above cases had been hit, but I didn't touch that code in this PR. Also, since I didn't touch `client.NewTFEAPI`, I did not update that case either.
image

product/consul.go
nwchandler: I added an item to the internal backlog (CORI-274).

client/apiclient.go
```diff
@@ -47,13 +50,82 @@ type TLSConfig struct {
...
+	tlsClientConfig, err := createTLSClientConfig(t)
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsClientConfig,
+		},
+	}
```
gulducat, line 63: the construction of the client and its transport here in NewAPIClient (for now at least) was exactly what I had in mind, glad it makes sense to ya here too :)

the only (vague) concern I have is that the other aspects of `http.DefaultTransport`, which `http.DefaultClient` uses, are excluded here.  I don't know to what extent they are actually important, but I think I would like to retain those defaults.

we can either copy-paste the whole [stdlib default `Transport` struct](https://cs.opensource.google/go/go/+/dev.boringcrypto.go1.17:src/net/http/transport.go;l=43-54) into here, or use it where it is and add to it, e.g.

```go
	transport := http.DefaultTransport.(*tls.Transport)
	transport.TLSClientConfig = tlsClientConfig
	client := &http.Client{
		Transport: transport,
	}
```

fwiw in the past @mkcp and I had agreed that copy-pasting the struct here is nice for explicitness and easier future flexibility if we want to change our defaults.

client/apiclient.go
```diff
@@ -47,13 +50,82 @@ type TLSConfig struct {
...
+
+	if foundClientCert {
+		// We use `GetClientCertificate` here because it works with Vault, along with other products. In Vault
+		// client authentication can use a different CA than the one used for Vault's certificate. However, it
+		// will indicate that its CA should be used when sending the client certificate request. If the client
+		// certificate does not share this CA, Go will fail with a "remote error: tls: bad certificate" error.
+		// By providing an override for `GetClientCertificate`, we ensure that we send the client certificate
+		// anytime one is requested, even if the CA does not match.
+		//
+		// See GitHub issue https://github.com/hashicorp/vault/issues/2946 for more context on why Vault
```
gulducat, line 121: I'm still not sure that we neeeed this for our client, since the issue seems to be more related to vault's tls cert _auth method_ than the more general tls verification we're looking to enable, but this is such interesting sleuthing!  and if it works for all of the products just the same, then that's just delightful.  thanks for the extra commentary about it here!

client/apiclient.go
```diff
@@ -47,13 +50,82 @@ type TLSConfig struct {
...
 	}
+
+	if t.CACert != "" || len(t.CACertBytes) != 0 || t.CAPath != "" {
+		rootConfig := &rootcerts.Config{
+			CAFile:        t.CACert,
+			CACertificate: t.CACertBytes,
+			CAPath:        t.CAPath,
+		}
+		if err := rootcerts.ConfigureTLS(tlsClientConfig, rootConfig); err != nil {
+			return nil, err
```
gulducat, line 88: personal preference: returning a `nil` pointer, even alongside an error, always makes me `panic` a little anticipatorily, so I usually prefer to return an empty thing 

```suggestion
			return &tls.Config{}, err
```

or however far along it has been built so far

```suggestion
			return tlsClientConfig, err
```


on the flip side, a full-on panic sure would let us know if we're not handling that error somewhere 😜

client/apiclient_test.go
```diff
@@ -11,6 +11,8 @@ import (
 type mockHTTP struct {
 	called []*http.Request
 	resp   string
+
+	Transport *http.Transport
```
gulducat, line 15: do you have any plans to add unit tests for our new code paths?  our functional tests should capture the actual, erm, functionality, but unit tests are quite nice too.

gulducat (changes requested):
nice! 🔑 🔓

my main change requests are to retain DefaultTransport settings, and either some more unit testing or a ticket to add more unit testing :)

client/apiclient.go
nwchandler: Yeah, it was hard for me to tell whether this would actually impact us or not. But, it seems to work just fine for our case, too, so I figured it would be better to avoid a potential future interop bug.

agent/agent.go
mkcp: Oh, that's my bad! code indentation threw me off a bit there, i thought that we were relying on error handling in the vault clause. here's what the diff view looked like for 63:

Screen Shot 2022-04-22 at 11 46 47 AM

client/apiclient.go
nwchandler: I have pushed a change that I think helps address this concern, ensuring we start from a `DefaultTransport` and attaching a TLS Client Config if needed. Per out-of-band discussion, it looks like the implementation of `DefaultTransport` changes between go 1.17 and 1.18, such that it feels like it could be a little brittle to copy/paste the struct definition into our code. If we find this to be worthwhile in the future, it may be worth re-investigating it then.

agent/agent.go
nwchandler: Oh, okay, cool! Glad I wasn't missing something! 😄

client/apiclient.go
nwchandler: From a quick scan through the code base, it looks like it is more common (today) to `return nil, err`. I wonder if this is something that we should chat about as a group and come up with a preferred style on? I can see the reasoning for the concern (the "panic" as it were! 😆 ), but it doesn't look like it's common in hcdiag today. (/cc @mkcp for thoughts too!)

client/apiclient_test.go
nwchandler: We have a work item on the internal backlog now to add some unit tests, and we should be getting to that issue very soon (prior to release).

gulducat (approved):
✅

PR #107, opened 2022-05-02
remove a.NumErrors and the fail-the-run check

mkcp:
We added a heuristic that if > 50% of the seekers fail in a run, the run is invalid and hcdiag exits status 1. As hcdiag has grown, having a heuristic by seeker count has felt arbitrary - we get useful info from seeker's failing sometimes. Crucially, when we made product runs concurrent, the NumErrors count was not changed to increment atomically.

In summary, we have a nondeterministically bugged outdated heuristic. Best to delete it.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -30,7 +30,6 @@ type Agent struct {
 	Start       time.Time `json:"started_at"`
 	End         time.Time `json:"ended_at"`
 	Duration    string    `json:"duration"`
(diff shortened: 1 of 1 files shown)


gulducat (approved):
I wonder if having error count in the manifest might be useful metadata in future, but I suppose ManifestSeekers having error/status fields would work for parsing out that info too if we want/need it.

mkcp:
Yeah - adding the status sum to manifest would be handy.

PR #109, opened 2022-05-05
add log/docker seeker

mkcp:
Added a seeker that checks if docker if available on the machine, and grabs logs from the given image name (currently called container, which docker has some fuzzy matching for). We also check if logs are empty in case nothing shows up in the time range.

While this is a good building block, there's going to be cases where there are many containers on a machine that are relevant and we need to grab logs for. One path forward is that we grab and write log files from _all_ containers that contain the product name.

### DIFF
--- a/product/consul.go
+++ b/product/consul.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/hashicorp/hcdiag/client"
 	s "github.com/hashicorp/hcdiag/seeker"
+	logs "github.com/hashicorp/hcdiag/seeker/log"
 )
 
 const (
@@ -43,6 +44,8 @@ func ConsulSeekers(tmpDir string, since, until time.Time) ([]*s.Seeker, error) {
 		s.NewHTTPer(api, "/v1/namespace"),
 		s.NewHTTPer(api, "/v1/status/leader"),
 		s.NewHTTPer(api, "/v1/status/peers"),
+
+		logs.NewDocker("consul", tmpDir, since, until),
 	}
 
 	// try to detect log location to copy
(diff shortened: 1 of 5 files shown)


product/consul.go
```diff
@@ -43,6 +44,8 @@ func ConsulSeekers(tmpDir string, since, until time.Time) ([]*s.Seeker, error) {
 		s.NewHTTPer(api, "/v1/namespace"),
 		s.NewHTTPer(api, "/v1/status/leader"),
 		s.NewHTTPer(api, "/v1/status/peers"),
+
+		logs.NewDocker("consul", tmpDir, since, until),
```
nwchandler, line 48: Will this assume that the running container is named "consul," or will Docker match based on image names?

mkcp: Docker CLI calls the param `CONTAINER` but will happily match the image name too.

seeker/log/docker.go
```diff
@@ -0,0 +1,99 @@
...
+
+// Docker allows logs to be retrieved for a docker container
+type Docker struct {
+	// Container is the name of the docker container to get logs from
+	Container string
+	// DestDir is the directory we will write the logs to
+	DestDir string
+	// Since marks the beginning of the time range to include logs
+	Since time.Time
+	// Until is not implemented yet, and is always "now"
```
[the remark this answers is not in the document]
mkcp: Deleted it on current revision

product/consul.go
nwchandler: Nice! I did not know that.

nwchandler (approved):

PR #112, opened 2022-05-10
Add APIConfig Struct

nwchandler:
This merge adds an `APIConfig` struct to collect the parameters required by an `APIClient`. The `NewAPIClient` function has been updated to use this new struct, and any callers of `NewAPIClient` have been updated to use an `APIConfig` structure.

`NewAPIClient` tests were updated to use this structure, but validation was not updated, so that we could verify that behavior remained consistent.

As a bit of cleanup, TLS config and setup were moved in the api client file, so that API concerns and TLS/HTTP Transport concerns were closer to similar code. Previously, they were interspersed.

### DIFF
--- a/client/apiclient.go
+++ b/client/apiclient.go
@@ -18,37 +18,6 @@ type HTTPClient interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
-// TLSConfig contains the parameters needed to configure TLS on the HTTP client
-// used to communicate with an API.
-type TLSConfig struct {
-	// CACert is the path to a PEM-encoded CA cert file to use to verify the
-	// server SSL certificate. It takes precedence over CACertBytes
-	// and CAPath.
-	CACert string
-
-	// CACertBytes is a PEM-encoded certificate or bundle. It takes precedence
-	// over CAPath.
-	CACertBytes []byte
-
-	// CAPath is the path to a directory of PEM-encoded CA cert files to verify
-	// the server SSL certificate.
-	CAPath string
-
-	// ClientCert is the path to the certificate for communication
-	ClientCert string
-
-	// ClientKey is the path to the private key for communication
-	ClientKey string
-
-	// TLSServerName, if set, is used to set the SNI host when connecting via
-	// TLS.
-	TLSServerName string
-
-	// Insecure enables or disables SSL verification. Setting to `true` is highly
-	// discouraged.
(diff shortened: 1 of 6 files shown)


nwchandler:
Comment to kick off functional tests. 😄

client/apiclient.go
```diff
@@ -18,41 +18,10 @@ type HTTPClient interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
-// TLSConfig contains the parameters needed to configure TLS on the HTTP client
```
nwchandler, line 21: FYI, the TLS stuff moved toward the bottom of the file. Sorry it makes this file kind of ugly to review.

client/apiclient.go
```diff
@@ -18,41 +18,10 @@ type HTTPClient interface {
...
-	TLSServerName string
-
-	// Insecure enables or disables SSL verification. Setting to `true` is highly
-	// discouraged.
-	Insecure bool
-}
-
 // NewAPIClient gets an API client for a product.
-func NewAPIClient(product, baseURL string, headers map[string]string, tlsConfig TLSConfig) (*APIClient, error) {
+func NewAPIClient(cfg APIConfig) (*APIClient, error) {
```
nwchandler, line 22: This is where the substantial changes in the file are.

client/apiclient.go
```diff
@@ -65,99 +34,26 @@ func NewAPIClient(product, baseURL string, headers map[string]string, tlsConfig
...
-		}
-		tlsClientConfig, err := tc.tlsConfigFunction(tc.tlsConfig)
-		if err != nil {
-			return nil, err
-		}
-		transport.TLSClientConfig = tlsClientConfig
-	}
-	return transport, nil
-}
+// APIConfig contains the configuration details for an APIClient.
```
nwchandler, line 42: This is also a primary change for the PR.

client/apiclient.go
```diff
@@ -229,3 +125,113 @@ func (c *APIClient) request(method string, path string, data []byte) (interface{
 
 	return iface, err
 }
+
```
nwchandler, line 128: Everything below here is just code that got moved from above, but there were no "new" changes.

mkcp (approved):
Just what we scoped out - appreciate the review notes on the core changes vs. some incidental moved functions, made reading through better!

PR #113, opened 2022-05-10
Allow APIClient Injection into Product Seeker Creation Functions

nwchandler:
This merge moves the responsibility of creating an `APIClient` out of the new product seeker functions and into the new product functions. Behavior should remain unchanged in this, but the separation allows for dependency injection and should enable easier maintenance in the future.

In addition, the TFE product setup was made more similar to the existing products, primarily by allowing errors to be returned from a couple of key functions (`client. NewTFEAPI` and `product. NewTFE`) and then bubbling those up accordingly.


mkcp (approved):
LGTM - appreciate the detail in making the TFE signature consistent and handling the error case!

PR #114, opened 2022-05-12
migrate journald to seeker/log with type implementing Runner and error types

mkcp:
Changes within:
- Migrate journald over to the new `seeker/log` pkg
- Implement it as a Runner type like other seekers, with `Journald.Run()` and `NewJournald()`
- Add error types for journald which pass back the result, error, and status rather than squashing it in place

The user-facing change is that users now get warned if systemctl is not enabled rather than those being debug-level command. Originally we decided to hide this error, but in migrating journalctl to match other seekers I decided to be explicit about error cases like others even if they don't need resolving.


mkcp:
Pushed up a draft to see the functional test result in CI

mkcp:
Screen Shot 2022-05-12 at 2 53 00 PM
The "systemctl not found" error state objectively got worse here. I think the best way to solve it is to parse out the `ExecError executable file not found` error from commander into an `CommanderNotFound` error, that we can then wrap with an actionable message for users in the journald seeker.

seeker/log/journald.go
```diff
@@ -0,0 +1,126 @@
...
+			since:   since,
+			until:   until,
+		},
+	}
+}
+
+// Run attempts to pull logs from journald via shell command, e.g.:
+// journalctl -x -u {name} --since '3 days ago' --no-pager > {destDir}/journald-{name}.log
+func (j Journald) Run() (interface{}, seeker.Status, error) {
+	// Check if systemd has a unit with the provided name, return a nil pointer
```
nwchandler, line 40: Nitpick: To me, the wording on this comment is just a little bit unclear. I made a suggestion that I think clarifies it a bit.

```suggestion
	// Check if systemd has a unit with the provided name, and return a nil pointer if not.
```

seeker/log/journald.go
```diff
@@ -0,0 +1,126 @@
...
+			command: cmd,
+			result:  fmt.Sprintf("%s", out),
+			err:     err,
+		}
+	}
+
+	// check if user is able to read messages
+	cmd = fmt.Sprintf("journalctl -n0 -u %s 2>&1 | grep -A10 'not seeing messages from other users'", j.service)
+	out, err = seeker.NewSheller(cmd).Run()
+	if err != nil {
```
nwchandler, line 56: I'm a little confused on this bit. If I'm reading this right, we'll return when `err != nil` (lines 56-58), and we'll also return when `err == nil` (lines 60-67). I think anything beyond here would be unreachable.

In the original file, it looks like the expectation was that `err` should NOT be nil, so I suspect perhaps lines 56-58 can just be removed.

nwchandler (changes requested):
I left one minor nitpick, but I had a question about the logic in one area.

seeker/log/journald.go
mkcp: hmmm - yeah that's a bug. fixing in this revision, thanks !

seeker/log/journald.go
mkcp: The "return a nil pointer" behavior this refers to is no longer valid so i'll just delete that section of the comment

nwchandler (approved):
👍 Looks good!

PR #116, opened 2022-05-16
add flags and agent config for DebugDuration and DebugInterval. 

mkcp:
This adds the CLI flags and starts passing the values through the agent. It's not done yet, because the values aren't applied in product seekers yet. It'll be easier to take params once some changes are applied via the TLSConfig refactors. So just pushing it up as a draft to get it up off my machine.

### DIFF
--- a/main.go
+++ b/main.go
@@ -160,11 +160,8 @@ func (f *Flags) parseFlags(args []string) error {
 	flags.BoolVar(&f.Version, "version", false, "Print the current version of hcdiag")
 	flags.DurationVar(&f.IncludeSince, "include-since", SeventyTwoHours, "Alias for -since, will be overridden if -since is also provided, usage examples: `72h`, `25m`, `45s`, `120h1m90s`")
 	flags.DurationVar(&f.Since, "since", SeventyTwoHours, "Collect information within this time. Takes a 'go-formatted' duration, usage examples: `72h`, `25m`, `45s`, `120h1m90s`")
-
-	// FIXME(mkcp): Figure out what is going on with backticked strings in the usage showing up next to the flag as the param.
-	flags.DurationVar(&f.DebugDuration, "debug-duration", TenSeconds, "`10s` How long to run product debug bundle commands. See: -duration in `vault debug`, `consul debug`, and `nomad operator debug`")
-	flags.DurationVar(&f.DebugInterval, "debug-interval", FiveSeconds, "`5s` How long metrics collection intervals in product debug commands last. See: -interval in `vault debug`, `consul debug`, and `nomad operator debug`")
-
+	flags.DurationVar(&f.DebugDuration, "debug-duration", TenSeconds, "How long to run product debug bundle commands. Provide a duration ex: `00h00m00s`. See: -duration in `vault debug`, `consul debug`, and `nomad operator debug`")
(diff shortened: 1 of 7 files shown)


mkcp:
Not sure exactly what lines lint is failing on here. there's no vars for `dur` and `interval` as far as I've been able to search for. Will debug more when I have time to do so, I don't want to merge in a branch with failing CI state.

main.go
```diff
@@ -16,7 +16,9 @@ import (
 const SemVer string = "0.3.0"
 
 // SeventyTwoHours represents the duration "72h" parsed in nanoseconds
-const SeventyTwoHours time.Duration = 259200000000000
+const SeventyTwoHours = 72 * time.Hour
```
nwchandler, line 19: I like this format much better! 😄

main.go
```diff
@@ -149,14 +157,16 @@ func (f *Flags) parseFlags(args []string) error {
 	flags.BoolVar(&f.TFE, "terraform-ent", false, "(Experimental) Run Terraform Enterprise diagnostics")
 	flags.BoolVar(&f.Vault, "vault", false, "Run Vault diagnostics")
 	flags.BoolVar(&f.AllProducts, "all", false, "DEPRECATED: Run all available product diagnostics")
+	flags.BoolVar(&f.Version, "version", false, "Print the current version of hcdiag")
+	flags.DurationVar(&f.IncludeSince, "include-since", SeventyTwoHours, "Alias for -since, will be overridden if -since is also provided, usage examples: `72h`, `25m`, `45s`, `120h1m90s`")
```
nwchandler, line 161: Should we replace the backticks in the usage strings with single quotes based on what we discovered with the flag help showing the first backticked string as the type of input? Or, would a bug on the backlog be better for this?

product/nomad.go
```diff
@@ -23,10 +24,20 @@ func NewNomad(cfg Config) (*Product, error) {
 		return nil, err
 	}
 
+	// Apply nomad duration and interval default if CLI is using global defaults
+	// NOTE(mkcp): This isn't ideal because we're using magic numbers here to match the CLI defaults. We could pass a
+	//  tuple or something in from the CLI that describes both the default and the user-set value... but i'm timeboxing this.
+	if cfg.DebugDuration == 10*time.Second {
+		cfg.DebugDuration = NomadDebugDuration
```
nwchandler, line 31: Not necessarily for this PR, but I wonder if it would be useful to pull some of the defaults out of the `main` package and into products. `main.TenSeconds` could become, for example, `product.DefaultDebugDuration`. `main.go` could then import this to insert into the help text, but we could then do some checks within product (such as here, with Nomad), where we could more easily check wither, for e.g., `if cfg.DebugDuration == DefaultDebugDuration {}`.

product/product.go
```diff
@@ -16,18 +16,15 @@ const (
 	Vault  = "vault"
 )
 
-const (
```
nwchandler, line 19: Heh, well, about that earlier comment about putting these into the `product` package... 😄 Maybe we SHOULD go ahead and do that?

nwchandler (approved):
Approved with comments for consideration either for this PR or for a future update.

main.go
mkcp: Definitely worth doing, I'm going to make a backlog issue for it so we can also ensure the input types/guides are all consistent and useful while we replace backticks too.

product/product.go
mkcp: Yeah - honestly it makes the most sense to keep the consts in the product pkg. I wondered a bit about how to make it available via the agent to keep the dependency graph cleaner... but this feels like the way to go.

PR #117, opened 2022-05-16
add host/iptables seeker

mkcp:
It would be nice if this were cross platform but this is a useful incremental step forward. We replace this with a "firewall" seeker in the future if needed.


nwchandler (approved):
LGTM!

PR #118, opened 2022-05-17
add EtcHosts seeker

mkcp:
I could go here or there about whether we ought to implement this with a copier. Reading it out of json is a bit of a pain without using, say, sed to convert the escaped newlines into real newline chars.

Here's the read from `jq . ...`:
Screen Shot 2022-05-17 at 3 54 38 PM

and jq piped into sed with `cat hcdiag-2022-05-17T225108Z/Results.json | jq . | sed 's/\\n/\n/g'`
Screen Shot 2022-05-17 at 3 54 46 PM


mkcp:
Neat - I didn't know there was a windows analog! I'll see about making a follow up issue for it.

PR #119, opened 2022-05-19
Add Support for Collecting /etc/fstab on Linux Systems

nwchandler:
This merge adds support for pulling filesystem table details, meaning the `/etc/fstab` file, from Linux systems. Other operating systems skip this check.


seeker/host/fstab_test.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			expected := tc.expected
+			result, status, err := tc.fstab.Run()
+			assert.Equal(t, expected.result, result)
+			assert.Equal(t, expected.status, status)
+			assert.Equal(t, expected.err, err)
```
mkcp, line 99: I dig the tests for these but it's recommended that you don't match error strings directly because it's brittle especially in tests. `assert.Error()` or `assert.NoError()` are sufficient and preferred for what we're testing here.

As an aside, in these cases you're redefining the same errors by making new errors with the same messages messages over. Using an error type would allow you to match the type via `assert.IsType(t, ...)`.  Under the hood that looks like: `reflect.TypeOf(err1) == reflect.TypeOf(err2)`

Here's a playground exploring error equality a bit more https://play.golang.org/p/-5d7lGkWWpU

.gitignore
```diff
@@ -22,3 +22,9 @@
 *.gz*
 *notes*
 .DS_Store*
+
+# GoLand files
```
mkcp, line 26: Would rather this arrives in a separate PR, but it's fine we can merge it in here 😸

mkcp (approved):
I shared some concerns about naming idioms, unnecessary state, and test brittleness. I'm going to optimistically approve as I don't want to block a merge later after revisions.

.gitignore
nwchandler: Heh, okay, thanks!

seeker/host/fstab_test.go
nwchandler: Good callout. I made some minor changes so that we could use `assert.Error()`.

product/host.go
```diff
@@ -27,5 +27,6 @@ func HostSeekers(os string) []*seeker.Seeker {
 		host.NewProcess(),
 		host.NewNetwork(),
 		host.NewIPTables(),
+		host.NewFstab(os),
```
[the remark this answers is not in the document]
nwchandler: I updated this. Thanks for the callout. I'm still getting used to the conventions with acronyms.

seeker/host/fstab.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+
+	"github.com/hashicorp/hcdiag/seeker"
+)
+
+var _ seeker.Runner = Fstab{}
+
+type Fstab struct {
+	os      string
+	sheller *seeker.Seeker
```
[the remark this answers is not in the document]
nwchandler: @mkcp and I had an offline conversation about this, but I'm capturing it here for posterity. The real reason I included these bits of state was so that it would be easier to test the behavior of the `FSTab` struct. Without doing something like this, I didn't know of a good way to mock out the underlying sheller that these seeker relies on.

PR #120, opened 2022-05-20
Revert "add flags and agent config for DebugDuration and DebugInterval. "

mkcp:
Reverts hashicorp/hcdiag#116

Added a bug, need to revert and reapply with the new patch.


mkcp:
Nevermind, rolling forward 🎢

PR #121, opened 2022-05-20
added a bug and forgot to check :facepalm:

mkcp:
Fix da bug (accidentally used Duration in nomad when i should've used interval)


nwchandler (approved):
👍

PR #122, opened 2022-05-20
add host/proc_file.go

mkcp:
The results entries aren't very readable without pretty-printing, but it works for now.

A better impl. may employ a list of file locations rather than commands, so a sheller (or catter!) can cat the files to results, and a copier copies the files over to the bundle.

Screen Shot 2022-05-20 at 1 35 48 PM
!


nwchandler (approved):
LGTM!

PR #123, opened 2022-05-20
Add Network Addresses & Interface Details

nwchandler:
This merge adds more detail to the host network seeker, switching out the built-in `net` package usage with `gopsutil`. The format of each interface will go from something like the following:

```json
{
  "Index": 6,
  "MTU": 1500,
  "Name": "en0",
  "HardwareAddr": "abcagv5S",
  "Flags": 19
}
```

to the following, after this merge:

```json
{
  "index": 6,
  "mtu": 1500,
  "name": "en0",
  "hardwareAddr": "aa:bb:cc:82:fe:52",
  "flags": [
    "up",
    "broadcast",
    "multicast"
  ],
  "addrs": [
    {
      "addr": "fe80::743f/64"
    },
    {
      "addr": "192.168.1.1/24"
    }
  ]
}
```

### DIFF
--- a/seeker/host/network.go
+++ b/seeker/host/network.go
@@ -1,10 +1,9 @@
 package host
 
 import (
-	"net"
-
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/hcdiag/seeker"
+	"github.com/shirou/gopsutil/v3/net"
 )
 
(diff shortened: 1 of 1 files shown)


mkcp (approved):
I am living for the +4-5 red diff here. Red diffs best diffs

PR #124, opened 2022-06-01
Refactor Version Code for Consistency with Other Products

nwchandler:
This PR updates the version code to be more consistent with other HashiCorp products. It is an internal change and should not have user impact. However, it is one step in the process that would allow us to release more frequently, and it also sets the stage to later include detailed version information in Manifest.json output files.


.go-version
```diff
@@ -0,0 +1 @@
+1.17.9
```
nwchandler, line 1: This is just prep for switching the release process, but it was a small change and generically related to versioning, so I included it.

version/version.go
```diff
@@ -0,0 +1,85 @@
...
+	// human updates under normal circumstances.
+	GitCommit string
+
+	// BuildDate is the date/time when the build was created. It is set by the build process and should not require
+	// human updates under normal circumstances.
+	BuildDate string
+)
+
+// VersionInfo is a container for version information.
+type VersionInfo struct {
```
nwchandler, line 32: This intermediate struct would allow us to add version details to manifest.json in the future.

version/version_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetVersion(t *testing.T) {
+	vi := GetVersion()
+	assert.Equal(t, Version, vi.Version)
+	assert.Equal(t, Prerelease, vi.Prerelease)
+}
+
+func TestVersionInfo_VersionNumber(t *testing.T) {
```
nwchandler, line 16: Tests for this seemed like they might be a little over the top, but I thought it would be worth some basic verification that we're producing a semantic version. In other words, `"<version core>[-<pre-release>][+<metadata]"`.

.go-version
mkcp:
```suggestion
1.17.9

```
Add newline to get rid of the little stopsign 😄

Makefile
```diff
@@ -42,5 +44,9 @@ clean: ## clean bin and bundle files
 	rm -rf bin/ hcdiag-*
 .PHONY: clean
 
+version: ## Show the version of the project
```
mkcp, line 47: Nice addition

version/version.go
```diff
@@ -0,0 +1,85 @@
+package version
+
+import (
+	"fmt"
+)
+
+var (
```
mkcp, line 7: These are all well commented but I'm not sure about about the purpose of the exported vars without any values set. Are they by the build process externally?

nwchandler: Good call. I had misremembered that a variable had to be exported in order for ldflags to override them, but that doesn't appear to be true (just did a little validation and read the documentation for `-X` in ldflags, which doesn't indicate a variable needs to be exported). I'll leave the comments, but I'll stop exporting these vars.

PR #127, opened 2022-06-03
Integrate GitHub Build Workflow

nwchandler:
This PR primarily aims to integrate the GitHub build workflow into the repo. Using prior work from HashiCorp release engineers, we now produce build artifacts for Windows, Linux (including deb and rpm packages), and Darwin on every push to GitHub on every branch, allowing quicker identification of build issues.

In the same spirit, unit tests and linting are now executed on every push as well.

To support the standardized build process, the Makefile has been updated so that the same process is used both in CI/CD pipelines as well as on developer machines as well.


mkcp (approved):
LGTM!

PR #132, opened 2022-06-13
Update Dependencies

nwchandler:
The dependency github.com/shirou/gopsutil/v3 is updated to address a build-time deprecation warning in MacOS Monterey (12.0) and beyond.

The dependency github.com/stretchr/testify has had a patch update applied.

The library github.com/hashicorp/go-multierror is no longer used and has been removed.


nwchandler:
For reference, the gopsutil issue is [described here](https://github.com/shirou/gopsutil/issues/1164).

mkcp (approved):

PR #133, opened 2022-06-13
Add a DryRun() method to agent

mkcp:
The original goal of this refactor was to swap out a different agent with different state for running the dryrun. However, i made the classic mistake of thinking returning an interface was a good idea -- go ensures it isn't! Instead, the simpler route is to reuse the same agent, but branch off early in the Run function to instead do DryRun things. 

The benefits of this is that we're no longer checking in with the agent state for the dryrun flag throughout the Run function. It all reads a lot more linear. The tradeoff is we a bit more code repetition now.

For the most part, dryrun steps don't rely on things to occur in the main run. However, there is a workaround for copiers containing the filepath in their ID. We writing a glob character * to Agent.tmpDir dryrun so that way those IDs are valid and can be copied into filters in a real run with a specific filepath.


agent/agent.go
```diff
@@ -130,6 +131,55 @@ func (a *Agent) Run() []error {
...
+	// glob "*" here is to support copy/paste of seeker identifiers
+	// from -dryrun output into select/exclude filters
+	a.tmpDir = "*"
+	a.l.Info("Would copy included files", "includes", a.Config.Includes)
+
+	// Running healthchecks for products. We don't want to stop if any fail though.
+	a.l.Info("Checking product availability")
+	if errProductChecks := a.CheckAvailable(); errProductChecks != nil {
+		errs = append(errs, errProductChecks)
+		a.l.Error("Product failed healthcheck. ensure setup steps are complete", "error", errProductChecks)
```
nwchandler, line 149: Perhaps we could provide a URL to where the user could find setup steps:
```suggestion
		a.l.Error("Product failed healthcheck. Ensure setup steps are complete (see https://github.com/hashicorp/hcdiag for prerequisites)", "error", errProductChecks)
```

agent/agent.go
```diff
@@ -130,6 +131,55 @@ func (a *Agent) Run() []error {
...
+	a.l.Info("Starting dry run")
+
+	// glob "*" here is to support copy/paste of seeker identifiers
+	// from -dryrun output into select/exclude filters
+	a.tmpDir = "*"
+	a.l.Info("Would copy included files", "includes", a.Config.Includes)
+
+	// Running healthchecks for products. We don't want to stop if any fail though.
+	a.l.Info("Checking product availability")
+	if errProductChecks := a.CheckAvailable(); errProductChecks != nil {
```
nwchandler, line 147: I wonder if dryrun should just check whether the required clients are installed using the products' ClientCheck commands instead of also running a healthcheck. I could see it being a bit unexpected to a user if dryrun made any outbound calls.

I know this is existing behavior, so this PR might not be the place to change it, but seeing the `CheckAvailable` call here reminded me of it.

agent/agent.go
```diff
@@ -130,6 +131,55 @@ func (a *Agent) Run() []error {
...
+		errs = append(errs, errProductChecks)
+		a.l.Error("Product failed healthcheck. ensure setup steps are complete", "error", errProductChecks)
+	}
+
+	// Create products and their seekers
+	a.l.Info("Gathering seekers for each product")
+	p, errProductSetup := a.Setup()
+	if errProductSetup != nil {
+		errs = append(errs, errProductSetup)
+		a.l.Error("Failed running Products", "error", errProductSetup)
```
nwchandler, line 157: Small suggestion on wording here - I could see someone getting a little confused about anything "running" (or failing to do so) when doing a dryrun. 
```suggestion
		a.l.Error("Failed gathering seekers for products", "error", errProductSetup)
```

nwchandler:
I also think we can get rid of the `TODO` on [agent/line 21](https://github.com/hashicorp/hcdiag/blob/main/agent/agent.go#L21) because with this PR, it's `TO-DONE`. 😄

agent/agent.go
mkcp: I agree that issuing http reqs does break the contract a little bit, even if it's helpful. Maybe this is our first usecase for multiple commands in hcdiag?

I could see adding:

`hcdiag validate`
`hcdiag dryrun`
`hcdiag run`

PR #136, opened 2022-06-14
migrate agent.runSet to product.Run

mkcp:
This PR aims to fix a strange abstraction jump where products store a list of seekers and have everything they need to run them, but instead the agent reached in and called seeker.Run() itself. This was a half-hearted attempt to use immutable data, but there's a better way we can accomplish the same thing. Instead, we have a product.Run() function that returns the immutable results, while all of the stateful stuff stays within the boundary of the methods operating on the product struct. Products also now need an application logger because that was previously read off of the agent.

Much better!

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -273,8 +273,6 @@ func (a *Agent) RunProducts() error {
 	wg.Add(len(a.products))
 
 	// NOTE(mkcp): Create a closure around runSet, writing results to the agent's result map, and wg.Done().
-	//   This is a little complex, but saves us duplication in the product loop. Maybe we extract this to a private
-	//   package function in the future?
 	run := func(wg *sync.WaitGroup, name string, product *product.Product) {
 		result := product.Run()
 
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -266,7 +266,6 @@ func (a *Agent) CopyIncludes() (err error) {
 }
 
 // RunProducts executes all seekers for this run.
-// TODO(mkcp): Migrate much of this functionality into the product package
 func (a *Agent) RunProducts() error {
 	// Set up our waitgroup to make sure we don't proceed until all products execute.
 	wg := sync.WaitGroup{}
@@ -276,8 +275,7 @@ func (a *Agent) RunProducts() error {
 	//   This is a little complex, but saves us duplication in the product loop. Maybe we extract this to a private
 	//   package function in the future?
 	run := func(wg *sync.WaitGroup, name string, product *product.Product) {
-		set := product.Seekers
-		result, err := a.runSet(name, set)
+		result, err := product.Run()
 		if err != nil {
 			a.l.Error("Error running seekers", "product", product, "error", err)
 		}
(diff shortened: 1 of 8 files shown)


product/product.go
```diff
@@ -33,10 +32,32 @@ type Config struct {
...
+	l        hclog.Logger
 	Name     string
 	Seekers  []*seeker.Seeker
 	Excludes []string
 	Selects  []string
+	Config   Config
+}
+
+// Run runs the seekers
+func (p *Product) Run() map[string]interface{} {
```
mkcp, line 44: I think it's time to make seekers run concurrently on the prod release, but do so with a context for tripping timeouts and storing retry counts. Retries and timeouts are most useful for network requests which might be flaky, and we'd probably want to make it configurable when seekers are custom defined.

agent/agent.go
```diff
@@ -264,11 +264,7 @@ func (a *Agent) RunProducts() error {
 	//   This is a little complex, but saves us duplication in the product loop. Maybe we extract this to a private
 	//   package function in the future?
 	run := func(wg *sync.WaitGroup, name string, product *product.Product) {
-		set := product.Seekers
```
nwchandler, line 267: It looks like the note on line 263 could maybe be removed now?

product/consul.go
```diff
@@ -26,8 +28,10 @@ func NewConsul(cfg Config) (*Product, error) {
 		return nil, err
 	}
 	return &Product{
+		l:       logger.Named("product"),
```
nwchandler, line 31: I'm not super familiar with the hclog library, but I wonder, if we're creating new named loggers for each product, should we also indicate the product type? For example, `"product.consul"`? This is just complete speculation/curiosity on my part.

product/product.go
```diff
@@ -33,10 +32,32 @@ type Config struct {
...
+	p.l.Info("Running seekers for", "product", p.Name)
```
nwchandler, line 45: This is probably a very small issue, but in theory, I think a person could create a Product without using one of the individual `New` commands, in which case `p.l` would be `nil`, and this would panic. I wonder if it would be worth skipping if `p.l` is nil or else assigning it a `hclog.Default()` if it were nil.

nwchandler (approved):
Looks good! I left a couple small thoughts for consideration, but this looks much better than before to me!

product/consul.go
mkcp: I tried it both ways. hclog convention is to name by subsystem which is more "product" than each individual product. We also print out the product name in logs down the line so the info is available. One way we could improve that though is storing the named logger `.With(product.Name)` which ensures that key is printed in every logline, much like we do now by hand in the line.

product/product.go
mkcp: That's a good point - it's true that it could panic. By convention the New... functions are there in go when the zero values aren't valid so not using it when they're is one would be a bug. The best place to fallback to hclog.Default would be if a logger weren't supplied in New... . Where this gets weird is that the new functions are per-product rather than using a NewProduct fn. I wouldn't expect this to be an issue except maybe for shortcuts in testing 🤔 

If it were anything other than logging i'd be inclined to guard against the nil, but something about multiple lines for logging feels _off_. Idk that's more an aesthetic concern - i'll keep an eye on it but otherwise will punt in this PR.

PR #138, opened 2022-06-15
Update Slack Notification Channel & Extend Integration Test Workflow Triggers

nwchandler:
The Slack notification channel in our release definition is updated to point to the hcdiag project channel, as opposed to a team channel.

Additionally, the functional test workflow has been updated for the following:
1. We can now trigger the workflow manually instead of only through automation;
2. Pushes to both main and any release branch will trigger the workflow;
3. The workflow is triggered when pull requests are opened, synchronized (changes pushed), or re-opened, as opposed to triggering on review comments.

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -1,11 +1,19 @@
 name: test-integration
 on:
+  # Allow manual runs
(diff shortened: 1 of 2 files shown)


mkcp (approved):
👍

PR #139, opened 2022-06-15
Standardize Agent Test Definitions

nwchandler:
This PR attempts to improve standardization among agent_test.go test definitions. Specifically, the following improvements are included:

1. Cases where Agent objects had been manually created have been replaced with calls to NewAgent()
2. Test Table definitions have been made more consistent with test tables in other parts of the code
3. Conditionals on test assertions have been replaced with more concise calls to functions from the testify/assert package
4. An unhandled deferred file cleanup error is now handled with a t.Log message


mkcp (approved):
Good stuff - glad we were able to make time for this

PR #140, opened 2022-06-16
Type seekers in results

mkcp:
We have a chance to lean on the type system a bit more to help the system make sense -- so here we swap out `interface{}` for `seeker.Seeker` copies in a few places in results. The result inside of a seeker is still `interface{}.


mkcp:
Forgot to rebase onto the test updates - will fix

nwchandler (approved):
👍

PR #141, opened 2022-06-16
Print Summary to Standard Out at end of Agent Run

nwchandler:
This merge adds a summary report to the end of an hcdiag agent run, with the intention of making it a bit easier for users to quickly see the overall status of seekers. The report is a simple table, which includes - for each product that was part of the run - a count of successful, failed, and unknown seekers, as well as the total number of seekers.

Unlike log messages, the report goes to Standard Out, meaning the run can be piped to an output file if desired.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -538,6 +538,22 @@ func ParseHCL(path string) (Config, error) {
 	return config, nil
 }
 
+func formatReportLine(cells ...string) string {
+	format := ""
+
+	// The coercion from the argument of type []string to type []interface is required for the later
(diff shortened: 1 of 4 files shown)


mkcp (approved):
I like how the table works out here, that's a nice touch. The all-caps header is good for writing to a file, but feels self-explanatory and not needed in logs. I don't feel too strongly about it, so your call if you want to leave it in.

nwchandler:
I removed the header. I agree, it was probably unnecessary for displaying in real-time to users.

PR #149, opened 2022-06-24
Split runners and ops

mkcp:
This is a follow up to renaming seekers in which we add some boundaries between Operations To Be Run (`Runner`s now) and the result of these operations `Op`.

A list of changes in this migration:
- Remove Runner field from Ops
- Store Runners on Product instead of Ops
- Change Runner.Run() to return an immutable Op struct copy and implement on all runners
  - Add an op.New() function to help make this easier.
  - Add Params field to Op and serialize Runner params onto it
- Add Runner.ID() method and implement on all runners
- Change New<Runner> functions to return the struct type it’s named after 
- Change Agent and Product reporting to use Ops and Runners where appropriate.

### DIFF
--- a/op/commander.go
+++ b/op/commander.go
@@ -5,8 +5,6 @@ import (
 	"fmt"
 	"os/exec"
 	"strings"
-
-	"github.com/hashicorp/hcdiag/util"
 )
 
 var _ Runner = Commander{}
@@ -36,16 +34,6 @@ func (c Commander) params() map[string]string {
 	}
 }
 
-func (c Commander) op(result interface{}, status Status, err error) Op {
-	return Op{
-		Identifier: c.ID(),
-		Result:     result,
-		Error:      err,
-		Status:     status,
-		Params:     util.RunnerParams(c),
-	}
-}
-
 // Run executes the Command
 func (c Commander) Run() Op {
 	var result interface{}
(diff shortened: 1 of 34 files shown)


changelog/149.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+op: store Runners on Products and change Runners interface to return an Op
```
nwchandler, line 2: Since these changelog files will wind up in release notes, what would you think about us standardizing on sentences for the descriptions? So, something like:

```suggestion
op: Store Runners on Products and change Runners interface to return an Op.
```

op/commander.go
```diff
@@ -37,7 +40,8 @@ func (c Commander) Run() (interface{}, Status, error) {
 	// Execute command
 	bts, err := exec.Command(cmd, args...).CombinedOutput()
 	if err != nil {
-		return string(bts), Unknown, CommandExecError{command: c.Command, format: c.format, err: err}
+		err1 := CommandExecError{command: c.command, format: c.format, err: err}
+		return New(c, string(bts), Unknown, err1)
```
nwchandler, line 44: I know it didn't change in this PR, but should we change `New` to `NewOp`? There are several different structs in package `op`, and I wonder if it would be easier to read if the name said what we're creating. (I say that as someone who had to look up what `New` was when reviewing this PR. 😄)

nwchandler (commented):
I had a couple small requests that I'd like to see in the PR (outdated comment and a changelog format update) and a more substantive one for consideration (change `New` to `NewOp` in the op package).

nwchandler (approved):
LGTM!

PR #150, opened 2022-06-27
Migrate runners to a runners pkg

mkcp:
This completes the code components of the seekers migration. Here we split the combined `op` package into `runner` and `op`.  This created a circular dependency, because Op accepted `Runner` as a param. Instead, we now accept the basic data types and collections at the function boundary rather than calling the methods on the calling object. You can see in the commit messages that there's a couple other fixes to the Results rendering.

Alongside with #147 and #149 , we introduce some breaking changes to the form of Results.json. The "runner" field is now effectively called "params" and renders more consistently across different `Runner` (prev. `seeker`) types. It will be omitted if a runner does does not take any params.


mkcp:
This includes the changes in #149 so the diff against the merge target, main, contains more than just the new commits. The diff should update once #149 is merged.

changelog/150.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:breaking
```
nwchandler, line 1: I wonder if we should also include an improvement note for this PR. There was also a lot done to improve the extensibility of the code going forward. I know it won't directly impact users, but it may warrant a blurb in the 0.4.0 changelog.

product/consul.go
```diff
@@ -7,9 +7,9 @@ import (
 	"github.com/hashicorp/go-hclog"
 
 	"github.com/hashicorp/hcdiag/client"
-	"github.com/hashicorp/hcdiag/op"
-	s "github.com/hashicorp/hcdiag/op"
-	logs "github.com/hashicorp/hcdiag/op/log"
+	"github.com/hashicorp/hcdiag/runner"
+	s "github.com/hashicorp/hcdiag/runner"
```
nwchandler, line 11: I know this was not part of the PR, but I see that we're importing `runner` directly as well as aliased to `s`. Could we get rid of the alias? It looks like it's only used on line 40.

runner/runner.go
```diff
@@ -0,0 +1,78 @@
...
+			newRunners = append(newRunners, r)
+		}
+	}
+	return newRunners, nil
+}
+
+// Params takes a Runner and returns a map of its public fields
+func Params(r Runner) map[string]interface{} {
+	var inInterface map[string]interface{}
+	inrec, _ := json.Marshal(&r)
```
nwchandler, line 75: Obligatory "Should we handle this error?" 😄

nwchandler (approved):
Looks good. I included a few small comments for consideration.

changelog/150.txt
mkcp: True - the changes ought to make contributing runners quite a bit easier!

product/consul.go
mkcp: There were a few places seeker was aliased - missed this one in the refactor, thanks!

runner/runner.go
mkcp: Ah, got caught copy/pasting 😅

PR #151, opened 2022-06-28
Upgrade to go 1.18 and use type union for HCLConfig

mkcp:
I figured that condensing how we process HostConfig and ProductConfig into one customRunners() function was a good usecase for go generics in 1.18. I bet we can cut down on duplication between the two switches too, so I'm uploading as a draft.


agent/agent.go
```diff
@@ -599,98 +599,96 @@ func formatReportLine(cells ...string) string {
...
-	for _, c := range cfg.Shells {
-		sheller := runner.NewSheller(c.Run)
-		runners = append(runners, sheller)
-	}
-
-	for _, g := range cfg.GETs {
-		runners = append(runners, host.NewGetter(g.Path))
-	}
+	switch cfg := any(config).(type) {
+	case *ProductConfig:
```
nwchandler, line 606: I notice that this function is now ~90 lines long and looks like it has several branches along the way. What if the new "genericized" function had the type switch, but then simply called out to non-generic functions based on the type? Essentially, the same two functions as before would exist and could be private/unexported functions, but this generic one would be the entrypoint? In prepping to split out a library, perhaps it could also be an exported function? It would still allow for a simpler interface for callers, I think, but allow us to test the sub-functions more easily.

nwchandler (changes requested):
The `.go-version` file will need to be updated to `1.18.3` (or another patch version of 1.18) in order to build correctly in the build system. This was a difference introduced with the new release tooling process.

I left a comment inline, but I also think there may be a way to use generics to simplify the experience of callers while also decomposing the `customRunners` func a bit.

agent/agent.go
mkcp: I ended up solving this problem by treating `customRunners()` as a reducer from `HCLConfig` to `[]Runner`. There's now mapping functions that correspond to each kind of RunnerConfig, like `[]ShellConfig`, and each applies the config structs to the corresponding `runner.New<Runner>()` constructor. This trims down `customRunners()` quite a bit and matches the shape of the data by destructuring each field in the `HCLConfig` types to their corresponding mapper.

nwchandler (approved):
Looks great to me! Much cleaner than the `customRunners` and `customHostRunners` from before. And, woohoo for generics! 😄

PR #152, opened 2022-06-29
Initial Support for Product Auto-Detection

nwchandler:
This PR adds support for product auto-detection. The hope is to make it simpler for users to get diagnostics on all installed products by simply running the `hcdiag` command. This is similar to the previous `all` flag, but with some basic checks in place to validate which products' CLIs are actually installed first.

### DIFF
--- a/changelog/147.txt
+++ b/changelog/147.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
-op: rename seeker to Op.
-`
+op: Rename seeker to Op.
+```
--- a/changelog/149.txt
+++ b/changelog/149.txt
@@ -1,3 +1,3 @@
 ```release-note:improvement
 op: Store Runners on Products and change Runners interface to return an Op.
-`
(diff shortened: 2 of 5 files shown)


changelog/147.txt
```diff
@@ -1,3 +1,3 @@
 ```release-note:improvement
-op: rename seeker to Op.
-`
+op: Rename seeker to Op.
```
mkcp, line 2: hehe, oopsie. Thanks 🙏

mkcp (approved):
I expected we'd need to solve this in the agent by rearranging the healthchecking but this is much simpler and composes well. Nice job

changelog/147.txt
nwchandler: Haha, I TOTALLY missed this in the other PRs!! I decided to see what our changelogs looked like, and there were some missing, which is the only reason I noticed this. 😄

PR #153, opened 2022-06-29
Deprecate the -all Command Line Flag

nwchandler:
The -all flag, which had previously been deprecated, is now removed. Documentation has been added to the README to provide details about the new auto-detection behavior, as it has taken the place of the -all flag.


mkcp (approved):
LGTM

PR #155, opened 2022-06-30
explcitly error if first CLI arg isn't a flag

mkcp:
Resolves #111 to cover the error case that the first arg provided is in subcommand form, not flag form. For subsequent args, flag parsing rules offer semantic errors.

PR #157, opened 2022-07-01
Allow Running with no Flags

nwchandler:
PR #155 addressed a bug when the first argument passed to hcdiag was not a flag. However, it also required at least flag to be present, or else it would panic. This PR addresses that limitation, so that hcdiag can be executed without specifying any flags.


mkcp (approved):
Whoops! Thank you for catching this

PR #159, opened 2022-07-14
Refactor config flow

mkcp:
Opening this up as a draft to get it off my laptop while I scope out more config changes.

This first step extracts the parsed-HCL out into an HCL pkg, which significantly cleans up some boundaries between agent and product.

I would like to continue by extracting runner generation from config off of the agent and into the runners pkg. However, I'm running into some circular dependencies because runner/host depends on runner, so host can't be imported into runner! Trying out a few different options now and seeing what fits best.


mkcp:
Migrated the customRunner creation to the hcl pkg, and added some end to end tests. This simplifies the agent a little bit more.

nwchandler (approved):
LGTM! 👍

PR #160, opened 2022-07-18
WIP: Capture more information about host processes

groovemonkey:
Instead of just grabbing process names, we'd like to capture name, PID, and PPID for all processes running on a host.

It might be nice to nest them under their PPID as well, to have a visual grouping that makes sense in Results.json.

**Jira Source**
- [Add PIDs to host/Process seeker](https://hashicorp.atlassian.net/browse/CORI-361)

### DIFF
--- a/runner/host/processes.go
+++ b/runner/host/processes.go
@@ -9,7 +9,12 @@ import (
 
 var _ runner.Runner = &Process{}
 
-type Process struct{}
+// Process represents a single OS Process
+type Process struct {
+	Name string `json:"name"`
+	PID  int    `json:"pid"`
+	PPID int    `json:"ppid"`
+}
 
 func (p Process) ID() string {
 	return "process"
@@ -22,10 +27,17 @@ func (p Process) Run() op.Op {
 		return op.New(p.ID(), processes, op.Fail, err, nil)
 	}
 
-	processInfo := make([]string, 0)
-	for eachProcess := range processes {
-		process := processes[eachProcess]
-		processInfo = append(processInfo, process.Executable())
+	// Maps parent PIDs to child processes
+	var processInfo = make(map[int][]Process)
+
+	for _, process := range processes {
+		proc := Process{
+			Name: process.Executable(),
+			PID:  process.Pid(),
+			PPID: process.PPid(),
+		}
+		// Append to slice of Process under this PPID
+		processInfo[proc.PPID] = append(processInfo[proc.PPID], proc)
 	}
 
 	return op.New(p.ID(), processInfo, op.Success, nil, nil)


runner/host/processes.go
```diff
@@ -22,10 +27,17 @@ func (p Process) Run() op.Op {
...
 	}
 
-	processInfo := make([]string, 0)
-	for eachProcess := range processes {
-		process := processes[eachProcess]
-		processInfo = append(processInfo, process.Executable())
+	// Maps parent PIDs to child processes
+	var processInfo = make(map[int][]Process)
+
+	for _, process := range processes {
```
groovemonkey, line 33: Would `for process := range processes {` be clearer or more idiomatic here?

groovemonkey:
I have a slightly wilder version of this code which re-processes the process list a second time, building a `map[string][]Process` that gives us processes nested under their Parent Process name, instead of Parent Process ID (PPID). That might be more readable for our users?

runner/host/processes.go
groovemonkey: Thank you! 🙇‍♂️ Really appreciate the explanation.

mkcp:
I like keying the map off the process name, but the process names don't provide the correct semantics. We would have to use a compound ID like `"{name}.{PID}"` to a guaranteed unique key that's also human readable.

runner/host/processes.go
```diff
@@ -22,10 +27,17 @@ func (p Process) Run() op.Op {
...
+		proc := Process{
+			Name: process.Executable(),
+			PID:  process.Pid(),
+			PPID: process.PPid(),
+		}
+		// Append to slice of Process under this PPID
+		processInfo[proc.PPID] = append(processInfo[proc.PPID], proc)
```
mkcp, line 40: This approach is a little bit between a tree representing parents and childen, and a flat list. What i mean by that, is when we keying and batch processes by their parent they show up multiple times in the readout, sometimes as the key of child processes, and other times as entries.

Handling arbitrary nesting like this gets pretty complicated, so lets instead render them as a flat map of processes with their PIDs, not PPIDs as their keys. For extra credit, a compound key like "{name}.{pid}" (or a pid,name tuple) could make it more readable. But there's edge cases to cover like what if there's a super long filename as the process. Maybe truncate the name to the first and last 5 letters like "pid-proce...name"?

Give it a try but it's ok to cut scope and just use the unique PID keys - the names are still available in the struct!

runner/host/processes.go
```diff
@@ -9,7 +9,12 @@ import (
 
 var _ runner.Runner = &Process{}
 
-type Process struct{}
+// Process represents a single OS Process
+type Process struct {
```
mkcp, line 13: Using the `Runner` type here to store one result from Run() isn't quite it yet. We set these fields as parameters when we create a `Runner`, so when we call the `process.Run()` later it has everything it needs. Put another way, this is the *Process Runner* not a data representation of a process's pid, ppid, and name.

Instead you'll need a data container to store and render the collection of processes that process.Run() generates. Try:

```
type proc struct {
        Name string `json:"name"`
	PID  int    `json:"pid"`
	PPID int    `json:"ppid"`
}
```

Finally, you can return a flat slice of results like `[]proc` - or `map[string]proc` if you decide to use a compound key as mentioned in other comments. There's no need to store anything on the `Process` struct because `Process.Run()` doesn't need any params. The results get returned on an immutable `Op{}` so no need to store those on `Process` either.

mkcp (changes requested):
Ensure that we're not storing any results on the Runner, and that the results are provided as a slice or use unique keys and you're good to go!

groovemonkey:
OK, I've added a bit more complexity while going for the extra-credit solution -- I'm now reprocessing the processlist together with the initial PPID -> []childproc map to resolve parent-process names.

Then, those process names are combined with their PIDs to make sure our final map is keyed on unique strings. They're also shortened in the way that Kit suggested; e.g.

`foobar` => `1-foobar`
`processfoobarname` => `7-proce...rname`
`boink333sdfsdfsdf122222222` => `15-boink...22222`

Results.json now looks like this:

```
root@hcdiagtest:/vagrant_data/hcdiag-2022-07-19T180254Z# zgrep -A 50 \"process\": Results.json
        "process": {
            "result": {
                "1-systemd": [
                    {
                        "name": "systemd-journal",
                        "pid": 356,
                        "ppid": 1
                    },
                    {
                        "name": "multipathd",
                        "pid": 399,
                        "ppid": 1
                    },
                    {
                        "name": "systemd-udevd",
                        "pid": 400,
                        "ppid": 1
                    },
...
```

runner/host/processes.go
```diff
@@ -22,11 +32,50 @@ func (p Process) Run() op.Op {
...
+			uniqueName := uniqueProcName(newProc)
+			namedProcessMap[uniqueName] = children
+		}
+	}
+
+	return op.New(p.ID(), namedProcessMap, op.Success, nil, nil)
+}
+
+// Creates a unique name from a proc's name and PID
+func uniqueProcName(process proc) string {
```
groovemonkey, line 72: Should this take a `*proc` instead?

nwchandler: If the state of `process` isn't being changed, a value type is probably better than a pointer. Pointers run the risk of side-effects, so if none are intended, then a value is usually best. (Exceptions can be made for very large objects, where passing a pointer might be enough of a memory improvement to warrant it. Although pointers also are more likely to wind up on the heap, which causes work for the garbage collector... Everything's a tradeoff. 😄 )

All that said, this MIGHT be a case where we should make this a method with a `proc` receiver. This function is unexported, so only used in the `runner/host` package, and it operates exclusively on a `proc`. So, perhaps it could become something like:

```go
func (p proc) uniqueName() string {
    pl := len(p.Name)
    // ... 
}
```

Then, lines 63-64 could become something like:

```go
namedProcessMap[newProc.uniqueName()] = children
```

groovemonkey: Thanks, that totally makes sense -- I was definitely thinking of optimizing away that tiny bit of memory usage for the value type, which is silly. And you're right, `uniqueProcName()` is very much "about" the proc, so I like the idea of giving it a proc receiver. I was trying to err on the side of avoiding things that looked like old OO habits 😄 .

Appreciate it!

groovemonkey:
Decided to go with something a lot more simple:
```
"process": {
            "result": [
                {
                    "name": "systemd",
                    "pid": 1,
                    "ppid": 0
                },
                {
                    "name": "kthreadd",
                    "pid": 2,
                    "ppid": 0
                },
```

mkcp (approved):
It's good to take the direct approach here with a flat list, which will match what folks expect when running ps in most cases! 👍

PR #162, opened 2022-07-20
Update issue templates

aliciaclark1066:


mkcp (approved):
We've discussed the template and tested the workflow in a test repo so: LGTM.

PR #166, opened 2022-07-25
Add log runners to HCL

mkcp:
This PR adds two runners: `log/journald` and `log/docker` to HCL and the README.


groovemonkey (approved):
This looks good -- a straightforward addition that adheres to the way all the other runners work. :shipit:

PR #167, opened 2022-07-26
Add "skip" status and colorize output appropriately

groovemonkey:
This change adds a new Op.Status named "skip," which gives us something beyond "success," "fail," or "unknown."

It also adds a colorization function in utils, which makes adding shell color codes to strings easy.


mkcp (approved):
Dave and I paired on this issue and added a new skip case to `log/journald` and made a revision and note to Error logging to ensure the test-functional workflow passes.

PR #169, opened 2022-07-27
Add redact blocks to HCL

mkcp:
This PR adds redact blocks to HCL runners and host+product blocks. If a host or product gets a redact then it will pass it through to every runner defined in HCL. This does not support HCL redact passthrough to a product/host's default runners, that will be implemented in a future PR.

### DIFF
--- a/changelog/169.txt
+++ b/changelog/169.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+hcl: Add redact blocks to product, host, and runners in HCL.
+```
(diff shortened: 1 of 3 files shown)


groovemonkey (approved):
OK, this seems cool -- basically we validate HCL config redactions, but stop short of adding them to the runners in their respective map functions. Looking forward to seeing this join up with #171 -- after that, I assume we'll create another PR to allow user-defined redactions to be added to our runners?

mkcp:
> OK, this seems cool -- basically we validate HCL config redactions, but stop short of adding them to the runners in their respective map functions. Looking forward to seeing this join up with #171 -- after that, I assume we'll create another PR to allow user-defined redactions to be added to our runners?

Yes, that's exactly it! I wanted to get the two dependencies in place before we start implementing in the runners themselves.

PR #171, opened 2022-07-27
Add Redact type, pkg, and validation

mkcp:
This PR adds the redact package, containing a single Redact type. In the process of building out the HCL, it became clear that, from a user's POV, a regex without anything special regexy behavior looks exactly like a string literal match. Therefore, I don't think we'll need multiple redact types and an interface. That means we'll have to go back and edit out the labels and validation in #169, but i figure we can handle that after both are merged, because this branch depends on the HCL additions.

This branch also sets up `Redact.Apply()` to take a writer and a reader, similar to `io.Copy()`. This prevents redact chaining, and requires that we copy the string we're redacting once per Apply. For multiple gigs of log files, that could result in a lot of extra copies per line. What we might be able to do as a workaround, is implement a `redact.ApplyAll([]Redact, r, w) err` which can accept many different redactors, each with their own replaces, and apply the highest precedence (default < global < product < runner) redacts first on a string.

### DIFF
--- a/changelog/169.txt
+++ b/changelog/169.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+hcl: Add redact blocks to product, host, and runners in HCL.
+```
--- a/changelog/171.txt
+++ b/changelog/171.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+redact: Add redact package and Redact type that allows a regex to be applied to a reader and written to a writer.
+```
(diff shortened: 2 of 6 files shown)


groovemonkey (approved):
This is great! I appreciate that you've gone with the minimal amount of complexity (a single Redact type, and doing the replace/redact in two steps for now).

PR #173, opened 2022-07-29
Add redact.ApplyMany

mkcp:
`redact.ApplyMany([]*Redact, io.Writer, io.Reader)` is similar to apply, but instead of being a method on a single redaction we take a collection of redactions and apply them in order to the result of Reader. This allows us to perform many redactions at once while only bringing any bytes to redact into memory once, and allows us to redact multi-gig log files if needed.

### DIFF
--- a/changelog/173.txt
+++ b/changelog/173.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+redact: Add ApplyMany function.
+```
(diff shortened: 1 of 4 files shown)


groovemonkey (approved):
Discussed with Kit via zoom -- I've seen how this gets used and it looks good!

PR #174, opened 2022-07-29
Redact HTTP and Shell runners

mkcp:
This PR adds type-specific functions: `redact.String(), redact.Bytes(), and redact.Files` that abstract away the readers and writers needed for `redact`'s `Redact.Apply` and `ApplyMany` functions. These provide a boundary for runners to pass and receive data back, instead of needing to be concerned with reader & writer behavior. The goal is to provide a toolkit of redactors that can meet the varied performance needs of different kinds of data.

In the PR we apply these to the HTTP and Shell runners. I decided to cut scope and solve commander and copier elsewhere because they each have more complex concerns.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -285,7 +285,7 @@ func (a *Agent) CopyIncludes() (err error) {
 		}
 
 		a.l.Debug("getting Copier", "path", f)
-		o := runner.NewCopier(f, dest, a.Config.Since, a.Config.Until).Run()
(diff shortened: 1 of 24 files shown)


groovemonkey (approved):
Discussed with Kit via Zoom -- this makes sense as a first crack at redactions.

groovemonkey (approved):
Thank you! The changes you've made look good, and are the missing link that I'll use to build agent redactions and redaction flow on.

PR #177, opened 2022-08-03
Redact command results

mkcp:
This PR introduces `redact.JSON()` and applies it and other redact functions to the output of Commander. Because redacts can return errors and we want to redact the results of errored commands, handling these cases in commander.Run gets a bit spaghettied - redact errors interleave with command errors. Anyway, here's an example of redacting from string values in JSON:

Command:
`VAULT_ADDR="http://127.0.0.1:8200" VAULT_FORMAT=json hcdiag run ./... -vault -config cfg/command-redacts.hcl`

HCL:
```
product "vault" {
  command {
    run = "vault status"
    format = "json"
    redact "regex" {
      match = "cluster"
    }
  }
}
```

```
"vault status": {
      "result": {
        "active_time": "0001-01-01T00:00:00Z",
        "cluster_id": "c8264088-5374-fd70-8041-d9652b0ef4c9",
        "cluster_name": "vault-<REDACTED>-f6c3d71f",
        "ha_enabled": false,
        "initialized": true,
        "migration": false,
        "n": 1,
        "nonce": "",
        "progress": 0,
        "recovery_seal": false,
        "sealed": false,
        "storage_type": "inmem",
        "t": 1,
        "type": "shamir",
        "version": "1.9.1"
      },
      "error": "",
      "status": "success",
      "params": {
        "command": "vault status",
        "format": "json",
        "redactions": [
          {
            "ID": "[6 178 212 185 27 92 158 170 140 32 161 194 112 249 91 60]",
            "replace": "<REDACTED>"
          }
        ]
      }
    }
```

### DIFF
--- a/redact/redact.go
+++ b/redact/redact.go
@@ -124,7 +124,7 @@ func JSON(a any, redactions []*Redact) (any, error) {
 		}
 		return r, nil
 	case []any:
-		r, err := redactArray(coll, redactions)
+		r, err := redactSlice(coll, redactions)
 		if err != nil {
 			return nil, err
 		}
@@ -134,7 +134,7 @@ func JSON(a any, redactions []*Redact) (any, error) {
 	}
 }
 
-func redactArray(a []any, redactions []*Redact) ([]any, error) {
+func redactSlice(a []any, redactions []*Redact) ([]any, error) {
 	for i, v := range a {
 		switch val := v.(type) {
 		case map[string]any:
@@ -144,7 +144,7 @@ func redactArray(a []any, redactions []*Redact) ([]any, error) {
 			}
 			a[i] = res
 		case []any:
-			res, err := redactArray(val, redactions)
+			res, err := redactSlice(val, redactions)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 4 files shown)


redact/redact.go
```diff
@@ -110,6 +114,82 @@ func Bytes(b []byte, redactions []*Redact) ([]byte, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		return r, nil
+	default:
+		return nil, fmt.Errorf("json collection to redact is neither a map nor an array, coll=%v", coll)
+	}
+}
+
+func redactArray(a []any, redactions []*Redact) ([]any, error) {
```
nwchandler, line 137: Sorry to be pedantic here, but should this be `redactSlice` actually?

nwchandler (approved):
I have a minor recommendation around the name of a newly added function, but otherwise, this looks good to me. :shipit:

redact/redact.go
mkcp: Might as well! We're a few steps past json.unmarshall at this point so it's not really a json array anymore.

PR #180, opened 2022-08-03
Add redactions to util/copyfiles and redact in Copier

mkcp:
Here we add line-by-line application of many redactions at once in `runner.Copier` and other places that use util/copyfiles.go. I ended up deleting the `redact.Files()` function because it didn't quite match the data boundary that the other redact functions have. Instead I utilized `redact.Bytes()` in our currently existing filecopy utils when redactions are passed in.

One benefit of this approach is that agent-level default and hcl redactions can be applied to `-include`ed files.

Here's an example of the redactor in action:

cli command: `go run ./... -autodetect=false -config cfg/copy-redacts.hcl `

HCL:
```
host {
  copy {
    path = "./tests/resources/example.log"
    redact "regex" {
      match = "kitpatella"
    }
    redact "regex" {
      match = "127.0.0.1"
    }
  }
}
```

Result in bundle:
Screen Shot 2022-08-03 at 3 31 39 PM

### DIFF
--- a/util/copyfiles.go
+++ b/util/copyfiles.go
@@ -1,11 +1,14 @@
 package util
 
 import (
+	"bufio"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/hashicorp/hcdiag/redact"
 
 	"github.com/hashicorp/go-hclog"
 )
@@ -13,7 +16,7 @@ import (
 const directoryPerms = 0755
 
 // CopyDir copies a directory and all of its contents into a target directory.
-func CopyDir(to, src string) error {
+func CopyDir(to, src string, redactions []*redact.Redact) error {
 	// get the absolute path, so we can remove it
 	// to avoid copying the entire directory structure into the dest
 	absPath, err := filepath.Abs(src)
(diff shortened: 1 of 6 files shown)


util/copyfiles.go
```diff
@@ -73,7 +76,25 @@ func CopyFile(to, src string) error {
...
 
-	// Write source contents to destination
-	_, err = io.Copy(w, r)
-	return err
+	if 0 < len(redactions) {
+		scanner := bufio.NewScanner(r)
+		// Scan, redact, and write each line of the src file
+		for scanner.Scan() {
+			bts := scanner.Bytes()
+			bts = append(bts, '\n')
```
groovemonkey, line 84: This tripped me up until I tested it myself. Didn't know that scanner.Scan() chews up newlines.

groovemonkey (approved):
This looks like a nice, clean implementation. :shipit:

nwchandler (commented):
The code looks good to me, but I did not see any new tests that include redactions on the copier. However, a new test resource file was added. Should we include a test to confirm that the redaction was applied properly?

util/copyfiles.go
mkcp: Yeah this was unintuitive - I had to break a file first to see it. The scanner impl. takes or applies a default split function  and like many split operations the delimiter is not in the result. I bet it's possible to work around this by passing our own split function.

mkcp:
> The code looks good to me, but I did not see any new tests that include redactions on the copier. However, a new test resource file was added. Should we include a test to confirm that the redaction was applied properly?

I spent some time updating the tests in copyfiles_tests.go but wasn't able to both keep it backwards compatible and get to meaningful redactions changes - so I'm going to ship a separate PR for those changes to avoid blocking this.

PR #183, opened 2022-08-04
Add Redaction Flow

groovemonkey:
This PR:

- adds agent redactions
- adds custom agent redactions, via a new `agent` HCL config block
- passes those agent redactions into Products and Hosts
- adds product and host redactions, via built-in defaults and the existing HCL config blocks
- ensures that HCL redactions are passed into (and used by) default runners
- ensures that more-specific redactions (e.g. HCL redaction on a custom runner) are applied BEFORE less-specific redactions (e.g. default agent redactions).


redact/default_agent_redactions.go
```diff
@@ -0,0 +1,36 @@
+package redact
+
+func GetDefaultAgentRedactions() ([]*Redact, error) {
```
mkcp, line 3: We should store this builder in the agent pkg, similar to how product defaults are in the product. `Agent` has to know about `Redact`s to be able to apply and pass them, but `redact` the pkg doesn't have to know about the Agent applies its Redact type. It's unlikely in this case, but in future designs: this would have the potential to run into a circular dependency if a default redact ever needed to grow to access `agent` pkg data. Essentially, the circular dep is already there at an abstract design level - but not yet in compiled types.

product/host.go
```diff
@@ -22,22 +26,37 @@ func NewHost(logger hclog.Logger, cfg Config, hcl2 *hcl.Host) (*Product, error)
...
+	builtInRunners := hostRunners(os, cfg.Redactions)
+	product.Runners = append(product.Runners, builtInRunners...)
+
 	return product, nil
 }
 
 // hostRunners generates a slice of runners to inspect the host.
-func hostRunners(os string) []runner.Runner {
+func hostRunners(os string, redactions []*redact.Redact) []runner.Runner {
+	// TODO(dcohen) Looks like none of these runners do redactions yet
```
mkcp, line 59: Yeah - host runners are so unlikely to reveal customer data that threading the redactions through are the last step. If any of these are higher order runners that implement commander or sheller then their impl. will already be taken care of, and we just need to pass in the redact data.

agent/agent.go
```diff
@@ -565,3 +573,22 @@ func formatReportLine(cells ...string) string {
 
 	return fmt.Sprintf(format, strValues...)
 }
+
+// setAgentRedactions mutates an Agent, setting agent-level redactions
+func setAgentRedactions(a *Agent) {
```
mkcp, line 578: Few comments here that would be blocking on a non-draft:

- Redactions failing for any reason should block operation of agent - we don't want to bundle up unredacted content. We ought to return any errors, not handle them w/ logging yet.
- Go convention would mean using a method with a pointer receiver here, rather than taking the pkg struct pointer as a first arg.
- We can take this design a step further and make this a referentially transparent function like:

```
package redact

// Flatten takes any number of slices of redacts and returns one slice containing all redacts in argument order
func Flatten(redacts ...[]*redact.Redact) []*redact.Redact { ... }
```

called as:

```
// NOTE: Would this work in agent.NewAgent()? we may not need any mutation
func (a *Agent) Setup() error {
...
hclRedacts, err := hcl.MapRedacts(a.Config.HCL.Agent.Redactions)
... (fail setup on err)
defaultRedacts, err := defaultRedacts()
... (fail setup on err)
redacts := redact.Flatten(hclRedacts, agentDefaults)
a.Redactions = redacts
// Note: double check the order on these i hope i did not get it backwards.
...
}

I think this will do a good job separating the "create a flat slice of redactions from two sources in a guaranteed order" concerns from the problems of sourcing multiple agent-specific redacts and storing those redacts on the agent.

It may even be possible to assign the redactions in agent.NewAgent without any mutation taking place!

File: product/host.go (hunk 2 above)
Comment (groovemonkey) [reply product/host.go:59]: Thanks, makes sense!

File: redact/default_agent_redactions.go (hunk 1 above)
Comment (groovemonkey) [reply redact/default_agent_redactions.go:3]: Thanks for pointing that out; I totally agree. This was the very first thing I implemented, before all the product defaults. I'll make it consistent with those.

Comment (groovemonkey) [reply redact/default_agent_redactions.go:3]: Just a quick follow up: I'd be happy to store this in a separate source file in the agent package, if it makes more sense to do that. Same for the default product-level redactions?

Not sure if that's cleaner than shoving this function into the already-long agent.go file, especially since this is more data than code.

File: agent/agent.go (hunk 3 above)
Comment (groovemonkey) [reply agent/agent.go:578]: This is a great idea, thank you! I like the Flatten semantics you've suggested -- will implement that.

Also happy to do it in NewAgent(); that's actually how I originally implemented it: https://github.com/hashicorp/hcdiag/compare/main...dcohen/redaction-exploration . I guess I got nervous about changing one constructor and making it potentially different from the others, but I agree that it's cleaner. Consider it done.

File: redact/default_agent_redactions.go (hunk 1 above)
Comment (mkcp) [reply redact/default_agent_redactions.go]: AFAIK there isn't much of any golang-level convention on when to split out files within a package. From a compiler view it all gets concatenated together and the order doesn't matter (just like how you can use a function above where it's defined in sourcecode). If `agent/redact.go` makes more sense for you, go for it!

File: agent/agent.go (hunk 3 above)
Comment (mkcp) [reply agent/agent.go]: Something that hadn't occurred to me even when I was refactoring a.Setup() that we may not even need the step. It's more or less doing part of the product & runner reification work that should be done in `agent.New()` and is left over from when the agent involved a lot more mutation throughout its runtime. The whole system gets a lot simpler if all of the agent data is truly fully initialized before it runs. And in the event we want to take in more config and runners during a.Run() we can offer a better API that's referentially transparent and can be unit tested.

Definitely not in scope for the redaction flow, but if you ever want to take on that refactor in the future it'll be an interesting one!

File: changelog/183.txt [hunk 4]
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+Add redaction flow.
```
nwchandler, line 2: We typically try and include a component at the beginning of the note, so that similar changes will be grouped together when the release notes are generated. I don't know if @mkcp has another preference, but perhaps this one could be `redaction`? So, it would become `redaction: Add redaction flow.` (I know that's a little redundant, but I think it would make the notes a little cleaner.)

hcl/hcl.go
```diff
@@ -320,7 +341,9 @@ func ProductsMap(products []*Product) map[string]*Product {
 	return m
 }
 
-func mapRedacts(redactions []Redact) ([]*redact.Redact, error) {
+// MapRedacts maps HCL redactions to "real" `redact.Redact`s
+// 	This is a public function so it can be used to map redactions from the Agent and Products
```
nwchandler, line 345: I don't know that the comment about the purpose of this being a public function adds much to understanding the code. Although we don't (yet) generate go docs for hcdiag, I also think this formatting might make its documentation look a little odd too. Just my $.02.

product/consul.go
```diff
@@ -69,9 +84,35 @@ func consulRunners(cfg Config, api *client.APIClient) ([]runner.Runner, error) {
...
 }
+
+// getDefaultConsulRedactions returns a slice of default redactions for this product
+func getDefaultConsulRedactions() []*redact.Redact {
+	redactions := []struct {
+		name    string
+		matcher string
+		replace string
+	}{
+		// {
```
nwchandler, line 101: Is this commented code intentional?

product/host.go
```diff
@@ -51,3 +69,29 @@ func hostRunners(os string) []runner.Runner {
...
 }
+
+// getDefaultHostRedactions returns a slice of default redactions for this product
+func getDefaultHostRedactions() []*redact.Redact {
+	redactions := []struct {
+		name    string
+		matcher string
+		replace string
+	}{
+		// {
```
nwchandler, line 80: Is this commented code intentional?

product/nomad.go
```diff
@@ -80,9 +95,35 @@ func nomadRunners(cfg Config, api *client.APIClient) ([]runner.Runner, error) {
...
 }
+
+// getDefaultNomadRedactions returns a slice of default redactions for this product
+func getDefaultNomadRedactions() []*redact.Redact {
+	redactions := []struct {
+		name    string
+		matcher string
+		replace string
+	}{
+		// {
```
nwchandler, line 112: Is this commented code intentional?

product/tfe.go
```diff
@@ -34,29 +41,62 @@ func NewTFE(logger hclog.Logger, cfg Config) (*Product, error) {
...
 }
+
+// getDefaultTFERedactions returns a slice of default redactions for this product
+func getDefaultTFERedactions() []*redact.Redact {
+	redactions := []struct {
+		name    string
+		matcher string
+		replace string
+	}{
+		// {
```
nwchandler, line 85: Is this commented code intentional?

product/vault.go
```diff
@@ -65,9 +80,35 @@ func vaultRunners(cfg Config, api *client.APIClient) ([]runner.Runner, error) {
...
 }
+
+// getDefaultVaultRedactions returns a slice of default redactions for this product
+func getDefaultVaultRedactions() []*redact.Redact {
+	redactions := []struct {
+		name    string
+		matcher string
+		replace string
+	}{
+		// {
```
nwchandler, line 97: Is this commented code intentional?

redact/redact_test.go
```diff
@@ -237,3 +238,104 @@ func TestJSON(t *testing.T) {
...
+			expect: []*Redact{multiRedact[0], multiRedact[1], singleRedact[0]},
+		},
+		{
+			name:   "Flatten should treat mixed args (nil slice, multiRedact, empty slice, singleRedact) correctly",
+			input:  [][]*Redact{nilSlice, multiRedact, emptySlice, singleRedact},
+			expect: []*Redact{multiRedact[0], multiRedact[1], singleRedact[0]},
+		},
+	}
+	// Run assertions
+	for _, tc := range tcs {
```
nwchandler, line 329: I see that this is the same pattern used in other tests in this file, but I'd propose a slight change - at least for consideration!

If we change lines 329-332 to this, instead:

```go
	for _, tc := range tcs {
		t.Run(tc.name, func(t *testing.T) {
			result := Flatten(tc.input...)
			assert.Equal(t, tc.expect, result, tc.name)
		})
	}
```

then, instead of having one test that loops through 13 cases, we wind up with 13 subtests that are part of one bigger test. That means output like this:

```
go test ./redact -run TestFlatten -v                                                                                                                                                                                      [12:20:19]
=== RUN   TestFlatten
--- PASS: TestFlatten (0.00s)
PASS
ok      github.com/hashicorp/hcdiag/redact      0.976s
```

becomes output like this:

```
go test ./redact -run TestFlatten -v                                                                                                                                                                                      [12:20:15]
=== RUN   TestFlatten
=== RUN   TestFlatten/Flatten_should_return_empty_redact_slice_for_nil_slice_input
=== RUN   TestFlatten/Flatten_should_treat_a_nil_slice_(first_arg)_correctly
=== RUN   TestFlatten/Flatten_should_treat_mixed_args_(multiRedact,_singleRedact,_nil_slice)_correctly
=== RUN   TestFlatten/Flatten_should_treat_mixed_args_(multiRedact,_nil_slice,_singleRedact)_correctly
=== RUN   TestFlatten/Flatten_should_treat_a_single_redact_input_correctly
=== RUN   TestFlatten/Flatten_should_treat_a_multi-redact_slice_input_correctly
=== RUN   TestFlatten/Flatten_should_treat_mixed-length_inputs_correctly_(1)
=== RUN   TestFlatten/Flatten_should_treat_mixed-length_inputs_correctly_(2)
=== RUN   TestFlatten/Flatten_should_return_empty_redact_slice_for_empty_redact_slice_input
=== RUN   TestFlatten/Flatten_should_treat_an_empty_slice_(first_arg)_correctly
=== RUN   TestFlatten/Flatten_should_treat_mixed_args_(multiRedact,_singleRedact,_empty_slice)_correctly
=== RUN   TestFlatten/Flatten_should_treat_mixed_args_(multiRedact,_empty_slice,_singleRedact)_correctly
=== RUN   TestFlatten/Flatten_should_treat_mixed_args_(nil_slice,_multiRedact,_empty_slice,_singleRedact)_correctly
--- PASS: TestFlatten (0.00s)
    --- PASS: TestFlatten/Flatten_should_return_empty_redact_slice_for_nil_slice_input (0.00s)
    --- PASS: TestFlatten/Flatten_should_treat_a_nil_slice_(first_arg)_correctly (0.00s)
    --- PASS: TestFlatten/Flatten_should_treat_mixed_args_(multiRedact,_singleRedact,_nil_slice)_correctly (0.00s)
    --- PASS: TestFlatten/Flatten_should_treat_mixed_args_(multiRedact,_nil_slice,_singleRedact)_correctly (0.00s)
    --- PASS: TestFlatten/Flatten_should_treat_a_single_redact_input_correctly (0.00s)
    --- PASS: TestFlatten/Flatten_should_treat_a_multi-redact_slice_input_correctly (0.00s)
    --- PASS: TestFlatten/Flatten_should_treat_mixed-length_inputs_correctly_(1) (0.00s)
    --- PASS: TestFlatten/Flatten_should_treat_mixed-length_inputs_correctly_(2) (0.00s)
    --- PASS: TestFlatten/Flatten_should_return_empty_redact_slice_for_empty_redact_slice_input (0.00s)
    --- PASS: TestFlatten/Flatten_should_treat_an_empty_slice_(first_arg)_correctly (0.00s)
    --- PASS: TestFlatten/Flatten_should_treat_mixed_args_(multiRedact,_singleRedact,_empty_slice)_correctly (0.00s)
    --- PASS: TestFlatten/Flatten_should_treat_mixed_args_(multiRedact,_empty_slice,_singleRedact)_correctly (0.00s)
    --- PASS: TestFlatten/Flatten_should_treat_mixed_args_(nil_slice,_multiRedact,_empty_slice,_singleRedact)_correctly (0.00s)
PASS
ok      github.com/hashicorp/hcdiag/redact      0.146s
```

In the case of failures, having the test name at the top of the output can be handy. Instead of this:

```
--- FAIL: TestFlatten (0.00s)
    redact_test.go:337: 
                Error Trace:    redact_test.go:337
                Error:          Not equal: 
```

we can get this:

```
--- FAIL: TestFlatten (0.00s)
    --- FAIL: TestFlatten/Flatten_should_treat_mixed_args_(multiRedact,_empty_slice,_singleRedact)_correctly (0.00s)
        redact_test.go:332: 
                Error Trace:    redact_test.go:332
                Error:          Not equal: 
 ...
```

Additionally, it allows running individual subtests, such as:

```shell
go test ./redact -run TestFlatten/"Flatten should treat mixed args \(multiRedact, empty slice, singleRedact\) correctly" -v
# the string is a regex so we have to escape the parentheses in this particular case
```

nwchandler (commented):
I don't yet have enough context on the redaction effort to really provide a lot of substantive commentary, but I did a quick "go" scan and had a few comments inline. 😄

changelog/183.txt
mkcp: I agree that it's worth being explicit here. Usually I add the pkg name which in this case would be `redact`, `agent`, `product`, and `hcl`. Then you can condense the pr description bullets that match each pkg in their changenote block.

agent/agent.go
```diff
@@ -565,3 +581,29 @@ func formatReportLine(cells ...string) string {
...
 }
+
+// GetDefaultAgentRedactions returns the default agent-level redactions that we ship with hcdiag
+func getDefaultAgentRedactions() []*redact.Redact {
+	redactions := []struct {
+		name    string
+		matcher string
+		replace string
+	}{
+		// {
```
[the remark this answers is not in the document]
groovemonkey: Yes, it's an example of how we'd add default redactions here, but I left it in a bunch of different files -- happy to clean that up if you and @mkcp prefer.

changelog/183.txt
groovemonkey: Oh, cool, so I would just add additional release-note blocks in the same changelog file for each pkg? Will do!

hcl/hcl.go
groovemonkey: That one was a note-to-self that got left; removed!

redact/redact_test.go
groovemonkey: That's so much better, and thank you for the detailed explanation! Adding this.

agent/agent.go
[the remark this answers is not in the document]
groovemonkey: I've removed all of these comments now.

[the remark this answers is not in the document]
nwchandler: The repetition across files gives me a bit of pause. It seems there's a pattern here of: empty slice of anonymous structs with name, matcher, replace fields; create a slice of redact.Redact using the struct slice as input; return the slice of redact.Redact. This appears to be happening across products.

So, I'm wondering if this isn't more of a concern of the redact package. Perhaps we could use a redactConfig struct (there's probably a better name - but it would have name, matcher, and replace). We could have a function to generate a redact.Redact from such a config. And, we could have a function that would take a slice of RedactConfig and return a slice of redact.Redact. Then, each of the products would have a (currently empty) slice of RedactConfig, it could pass that to the new redact slice function, and get back what it needs. I think that would deduplicate some code and could be beneficial in the future. Plus, if we wanted to add default redactors, we could just rely on the type of `[]redact.RedactConfig` to inform us what the shape of that input needed to be.

nwchandler (approved):
The latest changes really pulled things together in my opinion! Looks good to me!

product/consul.go
```diff
@@ -69,9 +84,20 @@ func consulRunners(cfg Config, api *client.APIClient) ([]runner.Runner, error) {
...
-		logCopier := runner.NewCopier(logPath, dest, cfg.Since, cfg.Until, nil)
+		logCopier := runner.NewCopier(logPath, dest, cfg.Since, cfg.Until, cfg.Redactions)
 		runners = append([]runner.Runner{logCopier}, runners...)
 	}
 
 	return runners, nil
 }
+
+// getDefaultConsulRedactions returns a slice of default redactions for this product
+func getDefaultConsulRedactions() []*redact.Redact {
```
mkcp: The `getDefault` part of the name isn't needed here or in other products where this is used. We can name it after the kind of data it returns, e.g. `consulRedactions`, just like `consulRunners`. In this way it's like storing the data in a var named `consulRedactions`, only we call the function to return it. We could keep default in the name, but then it would differ from the `<product>Runners` naming, so let's keep it similar.

hcl/hcl.go
```diff
@@ -99,9 +106,10 @@ func Parse(path string) (HCL, error) {
 // BuildRunners steps through the HCLConfig structs and maps each runner config type to the corresponding New<Runner> function.
 // All custom runners are reduced into a linear slice of runners and served back up to the product.
 // No runners are returned if any config is invalid.
-func BuildRunners[T Blocks](config T, tmpDir string, c *client.APIClient, since, until time.Time) ([]runner.Runner, error) {
+func BuildRunners[T Blocks](config T, tmpDir string, c *client.APIClient, since, until time.Time, redactions []*redact.Redact) ([]runner.Runner, error) {
```
mkcp, line 109: Not in scope for this PR but rather an observation that BuildRunners is still very complex. Despite my best effort to improve it last month we should probably get rid of the union type param and have separate functions for Host (later Platform) and Product - the repetition is still there with the type switch.

product/host.go
```diff
@@ -51,3 +69,14 @@ func hostRunners(os string) []runner.Runner {
 		host.NewFSTab(os),
 	}
 }
+
+// getDefaultHostRedactions returns a slice of default redactions for this product
+func getDefaultHostRedactions() []*redact.Redact {
```
mkcp: `s/getDefaultH/h` per comment above.

product/nomad.go
```diff
@@ -80,9 +95,20 @@ func nomadRunners(cfg Config, api *client.APIClient) ([]runner.Runner, error) {
...
-		logCopier := runner.NewCopier(logPath, dest, cfg.Since, cfg.Until, nil)
+		logCopier := runner.NewCopier(logPath, dest, cfg.Since, cfg.Until, cfg.Redactions)
 		runners = append([]runner.Runner{logCopier}, runners...)
 	}
 
 	return runners, nil
 }
+
+// getDefaultNomadRedactions returns a slice of default redactions for this product
+func getDefaultNomadRedactions() []*redact.Redact {
```
mkcp: `s/getDefaultN/n` per comment above.

product/tfe.go
```diff
@@ -34,29 +41,47 @@ func NewTFE(logger hclog.Logger, cfg Config) (*Product, error) {
...
 
-		runner.NewSheller("getenforce", nil),
-		runner.NewSheller("env | grep -i proxy", nil),
+		runner.NewSheller("getenforce", cfg.Redactions),
+		runner.NewSheller("env | grep -i proxy", cfg.Redactions),
 	}, nil
 }
+
+// getDefaultTFERedactions returns a slice of default redactions for this product
+func getDefaultTFERedactions() []*redact.Redact {
```
mkcp: `s/getDefaultTFE/tfe` per comment above.

product/vault.go
```diff
@@ -65,9 +80,20 @@ func vaultRunners(cfg Config, api *client.APIClient) ([]runner.Runner, error) {
...
-		logCopier := runner.NewCopier(logPath, dest, cfg.Since, cfg.Until, nil)
+		logCopier := runner.NewCopier(logPath, dest, cfg.Since, cfg.Until, cfg.Redactions)
 		runners = append([]runner.Runner{logCopier}, runners...)
 	}
 
 	return runners, nil
 }
+
+// getDefaultVaultRedactions returns a slice of default redactions for this product
+func getDefaultVaultRedactions() []*redact.Redact {
```
mkcp: `s/getDefaultV/v` per comment above.

redact/redact.go
```diff
@@ -187,3 +196,25 @@ func redactMap(m map[string]any, redactions []*Redact) (map[string]any, error) {
...
+func Flatten(redacts ...[]*Redact) []*Redact {
+	flattened := make([]*Redact, 0)
+	for _, rslice := range redacts {
+		flattened = append(flattened, rslice...)
+	}
+	return flattened
+}
+
+// MapNew takes a slice of RedactConfig and return a slice of redact.Redact
+func MapNew(configs []Config) ([]*Redact, error) {
```
mkcp, line 210: Crispy 👍

mkcp (approved):
We've talked about this a bunch already, so while there's a couple of nit comments and Future Stuff it's otherwise good to go.

product/consul.go
groovemonkey: That totally makes sense. Fixed!

PR #185, opened 2022-08-05
Update copyfiles tests to use test tables

mkcp:
Continues where #180 left off to update `util/copyfiles_test.go` and test it with redactions applied. I learned while writing these changes that we don't actually use CopyDir outside of Copier.Run() -- everywhere we copy files it's done with our Copier. Didn't quite get to applying redactions yet, because the setupFiles part still needs to be reworked. At least this way we have test tables and a few empty-params cases handled. More incremental progress to come in the future!


groovemonkey (approved):
Thanks, I'm going to use this as an example of what we want our tests to look like.

PR #186, opened 2022-08-05
add TFE runners to gather env data

anniehedgpeth:
[Jira - TF-449](https://hashicorp.atlassian.net/browse/TF-449)
This branch adds some easy checks to the TFE instance for the sake of data collection across customers' environments per this [PRD](https://docs.google.com/document/d/1yCoD2EAIoi6w89IJ4aefAAAssDd75fQRzffvT10H5xM/edit?usp=sharing) and its associated [RFC](https://docs.google.com/document/d/1rq6wvO8lfYya30zIZUKI7bmofDvhkyUkXt-YrwJe3SU/edit?usp=sharing).

**Docker version**
`docker -v` returns only the version on one line, like `Docker version 20.10.7, build f0df350`, as opposed to `docker version`, which returns more info than is needed.

**TFE Version**
`replicatedctl app status` returns the TFE sequence as well as the status of the TFE application.

**Mounted Disk - Memory**
For Mounted Disk mode, we'd like to know the amount of memory allocated to the disk used by TFE, however, there's not a way to specify whether or not it's mounted disk mode before running `lsblk` and still use the `commander` seeker. And I'd like to use the `commander` as opposed to `sheller` seeker so that I can get `json` back instead of a string.

**SELinux Enabled**
The `getenforce` sheller is meant to detect if SELinux is enabled. `getenforce` comes from `libselinux-utils` which, if SELinux is installed, should mean it's installed, too. However, if the `getenforce` command does not exist, then `hcdiag` will throw out the warning...
```
2022-08-04T22:26:03.671Z [WARN]  hcdiag.product: result: runner=getenforce status=unknown
  result=
  | /bin/bash: getenforce: command not found
   error="exec error, command=getenforce, error=exit status 127"
```
...which signals to us that SELinux is not enabled. 

**Proxy Used**
Likewise, if a `HTTP_PROXY` or `HTTPS_PROXY` environment variable does not exist, then we can assume that a proxy is not in use, so an empty string or newline value is acceptable.


product/tfe.go
```diff
@@ -51,5 +51,12 @@ func tfeRunners(cfg Config, api *client.APIClient) ([]runner.Runner, error) {
 		runner.NewHTTPer(api, "/api/v2/admin/twilio-settings", nil),
 		// page size 1 because we only actually care about total workspace count in the `meta` field
 		runner.NewHTTPer(api, "/api/v2/admin/workspaces?page[size]=1", nil),
+
+		runner.NewCommander("docker -v", "string", nil),
+		runner.NewCommander("replicatedctl app status --output json", "json", nil),
+		runner.NewCommander("lsblk --json", "json", nil),
+
+		runner.NewSheller("getenforce", nil),
+		runner.NewSheller("echo $proxy_ip", nil),
```
sudomateo, line 60: Do we expect `proxy_ip` to already be set/exported by something else? If so, what is that? I would expect us to look for `HTTP_PROXY` or `NO_PROXY` env vars instead.

anniehedgpeth: Oh whoops, you're right. That was an oversight. Should I look for both `HTTP_PROXY` and `HTTPS_PROXY`?

mkcp (approved):
@anniehedgpeth Great stuff on this PR - we're happy to continue accepting more improvements on TFE's default runners as you need them. Let me know when you're ready to merge and I or another team member will click the button 👍

anniehedgpeth:
> @anniehedgpeth Great stuff on this PR - we're happy to continue accepting more improvements on TFE's default runners as you need them. Let me know when you're ready to merge and I or another team member will click the button 👍

Thanks, @mkcp! It's good to go now (after this tiny update). Feel free to merge whenever you're ready for it.

product/tfe.go
sudomateo: We talked about this out of band. Landed on `env | grep -i proxy`. Marking this resolved.

PR #189, opened 2022-08-11
Accept and pass redaction config through in higher order runners

groovemonkey:
This PR adds redaction pass-through for higher-order host runners.

docker-log and journald-log reification now also takes redactions into account -- custom redactions can be added with an HCL block as shown in this example docker-log runner:
```
  docker-log {
    container = "arch-testcontainer"
    redact "literal" {
      match = "teststring"
      replace = "HCL: product 'consul' -> dockerlog -> redact"
    }
  }
```

### DIFF
--- a/runner/log/journald.go
+++ b/runner/log/journald.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/hcdiag/op"
+	"github.com/hashicorp/hcdiag/redact"
 
 	"github.com/hashicorp/hcdiag/runner"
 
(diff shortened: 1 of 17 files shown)


runner/log/journald_test.go
```diff
@@ -4,54 +4,60 @@ import (
...
-			since:   time.Time{},
-			until:   time.Date(2, 1, 1, 1, 1, 1, 1, &time.Location{}),
-			expect:  "journalctl -x -u nomad --until \"0002-01-01 01:01:01\" --no-pager > /testing/test/journald-nomad.log",
+			desc:       "With until but not since",
+			name:       "nomad",
+			destDir:    "/testing/test",
+			since:      time.Time{},
+			until:      time.Date(2, 1, 1, 1, 1, 1, 1, &time.Location{}),
+			expect:     "journalctl -x -u nomad --until \"0002-01-01 01:01:01\" --no-pager > /testing/test/journald-nomad.log",
+			redactions: nil,
```
mkcp, line 46: Small nit, but you don't need to pass in nil for redacts in these struct - they're already set nil so it's fine to elide it.

runner/log/journald.go
```diff
@@ -40,7 +43,7 @@ func (j Journald) ID() string {
 // Run attempts to pull logs from journald via shell command, e.g.:
 // journalctl -x -u {name} --since '3 days ago' --no-pager > {destDir}/journald-{name}.log
 func (j Journald) Run() op.Op {
-	o := runner.NewSheller("journalctl --version", nil).Run()
+	o := runner.NewSheller("journalctl --version", j.Redactions).Run()
```
mkcp, line 46: I was a bit nervous about redacting intermediate runners that aren't in the last Op that get returned, but I see now that ofc we do return the results when it errors. We might as well let the runner redact itself like we have here - it's unlikely to cause issues in the current log runners cases.

mkcp (approved):
LGTM

runner/log/journald.go
groovemonkey: Yeah, at first I thought the same thing and planned on writing an Op-redacter that we could just use a whole bunch (`redact.Op(op.Op)` or perhaps the same thing in the Op package). Thankfully this simple implementation seems fine for now.

PR #190, opened 2022-08-11
swap deprecated io/ioutil pkg for io and os

groovemonkey:
Linter happiness is the purest and most genuine happiness, and we should respect that.


dgregoire:
> Linter happiness is the purest and most genuine happiness, and we should respect that.

I wish to second this thought

mkcp (approved):
LGTM 👍

PR #191, opened 2022-08-11
DON'T MERGE: Initial spike on go library redaction

nwchandler:
Just creating a draft to solicit some feedback on this approach...

### DIFF
--- a/redact/types.go
+++ b/redact/types.go
@@ -1,0 +1,40 @@
+package redact
+
+import (
+	"encoding/json"
+)
+
+type RedactedString struct {
+	inputString string
+	redactions  []*Redact
+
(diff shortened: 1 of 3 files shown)


redact/types.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"encoding/json"
+)
+
+type RedactedString struct {
+	inputString string
+	redactions  []*Redact
+
+	redactedString string
```
nwchandler, line 11: Good call by the linter, but I had thought about using this to store the redacted string and then, if it's not empty, just displaying it out of the string method. I thought this could potentially save re-running redactors unnecessarily.

PR #194, opened 2022-08-12
fix sync bug

aliciaclark1066:


tehut (approved):
👍  Approving since most of the hc-diag SMEs are offline & @aliciaclark1066 is the JIRA integration SME

PR #195, opened 2022-08-16
Remove the unused function redact.Apply(), rename redact.ApplyMany

groovemonkey:
Does what it says on the tin: removes redact.Apply and renames redact.ApplyMany to redact.Apply.


nwchandler (approved):
LGTM!

PR #197, opened 2022-08-16
Add RegEx Capture Group Test Cases to Redaction Unit Tests

nwchandler:
This adds a few test cases for various uses of capture groups used for more complex replacement patterns than string literals. The functionality is provided "for free" by the Go standard library, but the tests intend to both confirm that they can be used as expected as well as to provide some examples of how this would be done.


groovemonkey (approved):
Nice, this is good to have.

PR #201, opened 2022-08-19
DockerNotFoundError should not be an error

groovemonkey:
My first crack at this just fixes the symptom we talked about yesterday: it makes a DockerNotFoundError a skip instead of a failure.

Another approach would be to remove the docker runner from the default product runners (except for TFE, where docker is expected). Interested to hear what you have to say.


mkcp (approved):
Logically it makes sense to skip the runner if docker isn't installed. If it needs to succeed for sequential runners in the future then we can add an optional attribute.

PR #203, opened 2022-08-22
BUG: do not serialize hcl.Redact.Match into json for the Manifest

mkcp:
This is a crucial bug that we caught in the RC: matchers that are provided in HCL currently get serialized into JSON because, while redact.Redact uses a private field and is not serialized, hcl.Redact is all public and therefore requires json annotations. I'll be making a follow up PR to correctly format all of the HCL fields in manifest to match their hcl representation.


nwchandler (approved):
LGTM!

groovemonkey (approved):
Whew! Nice catch.

PR #204, opened 2022-08-22
HCL Cleanup - redact literal label and json tagging

mkcp:
There's two scopes of changes in this PR:
First we fix a bug where we were checking for a redaction feature we cut but will add in the future: literals. This has been commented out with a TODO. Now when "literal" is provided in config, we error out the default case.

Next, we ensure the HCL structs have proper lowercased json tagging. We also set `omitempty` for every block / collection

When no config is provided, we used to:
Screen Shot 2022-08-22 at 1 52 55 PM

Now we:
Screen Shot 2022-08-22 at 1 52 46 PM

And so on, with each nested block being omitted if it is empty.

### DIFF
--- a/hcl/hcl.go
+++ b/hcl/hcl.go
@@ -15,9 +15,9 @@ import (
 )
 
 type HCL struct {
-	Host     *Host      `hcl:"host,block" json:"host"`
-	Products []*Product `hcl:"product,block" json:"products"`
-	Agent    *Agent     `hcl:"agent,block" json:"agent"`
+	Host     *Host      `hcl:"host,block" json:"host,omitempty"`
+	Products []*Product `hcl:"product,block" json:"products,omitempty"`
+	Agent    *Agent     `hcl:"agent,block" json:"agent,omitempty"`
 }
 
 type Blocks interface {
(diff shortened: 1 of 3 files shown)


hcl/hcl.go
```diff
@@ -26,73 +26,73 @@ type Blocks interface {
...
+	Excludes     []string      `hcl:"excludes,optional" json:"excludes,omitempty"`
+	Selects      []string      `hcl:"selects,optional" json:"selects,omitempty"`
+	Redactions   []Redact      `hcl:"redact,block" json:"redactions,omitempty"`
 }
 
 type Redact struct {
-	Label   string `hcl:"name,label"`
-	ID      string `hcl:"id,optional"`
+	Label   string `hcl:"name,label" json:"label"`
+	ID      string `hcl:"id,optional" json:"ID"`
```
nwchandler, line 59: Was this ID supposed to be lowercase?

nwchandler (approved):
Looks good, although I left a comment to confirm whether something was intentional.

groovemonkey (approved):

hcl/hcl.go
mkcp: I went with a capitalized acronym here, and JSON style guides conveniently seem to not mention acronym capitalization. I talked this over with a resident json/js expert (my partner) and he also said lowercase it. It makes it consistent with all the other fields, so let's go with that.

PR #207, opened 2022-08-23
Implement Default Redactions

nwchandler:


mkcp (approved):
Good stuff - esp. glad to see hardware addresses are no longer being collected 👍

PR #215, opened 2022-08-25
BUG: Ensure JSON HTTPer results don't get squashed into a flat string

mkcp:
Resolves an implementation bug where I converted an HTTP body into a string before redacting it as a string. It turns out that the apiclient get methods actually request JSON and unmarshal it into map[string]any. So squashing it into a string was bad news and broke some downstream results marshaling.

There's a larger design issue with the boundary between apiclient and runners. Over time we've been migrating external operations in hcdiag into the runners themselves, and apiclient exposes methods that are used by both runners and for querying products directly for their log locations in config. Migrating this pattern into the runner will lead to a more portable solution that composes well with other runner features.

### DIFF
--- a/changelog/215.txt
+++ b/changelog/215.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+runner: Fixed a redaction and reporting issue where HTTPer incorrectly flattened and redacted JSON responses as strings.
+```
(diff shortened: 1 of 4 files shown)


nwchandler (approved):
LGTM!

PR #216, opened 2022-08-25
Update release/v0.4.x branch with HTTPer bugfix

mkcp:


nwchandler (approved):
Looks good!

mkcp:
Github might be experiencing some downtime - the macos integration tests are spinning waiting to be created and i get 500s when I try to check. I'm reasonably confident that this will work given that macos passed with the same commits we're cherrypicking here.

PR #222, opened 2022-09-01
Incorporate mitchellh/cli Library into hcdiag

nwchandler:
This is the initial inclusion of the mitchellh/cli library into hcdiag. Going alongside this, it also makes some related changes that intend to smooth the process of adding additional subcommands in the future.

The main.go file is now primarily used to set up and execute the CLI. Flag parsing and agent instantiation have moved into the `cmd/run` package. (This package may be renamed at some point, but it is the holder of the previous setup around diagnostic collection.)

Executing hcdiag without any subcommands will execute a diagnostic run, matching current (0.4.0) behavior, however the `run` subcommand is preferred. The help text for the application indicates that this behavior will change in the future, and users should begin to use the `run` subcommand.

The version display has become its own subcommand. However, the `--version` flag will also still work. This behavior is more consistent with other HashiCorp products, which commonly have their own `version` subcommand.


cmd/help/usage.go
```diff
@@ -0,0 +1,75 @@
+package help
```
nwchandler, line 1: FYI, I borrowed liberally from consul in this file. 😄

mkcp (commented):
Good details in implementation, thinking through the abstractions, and making sure things are well commented. The addition of mainhelp text is especially good for guiding users. 

Before we're ready to merge, I have some pkg layout revisions change requests that will make the impl. Go-idiomatic and consistent with other hashicorp products.

In Go, typically you don't see packages that are defined to only contain other packages - that would likely trip up future maintainers. I recommend lifting the contents of `help` and `returns` up to the scope of `cmd`. To account for the shared constants defined here and avoid circular deps when referring to it in `cmd` we'll have to pull up `run` and `version` as well. I believe that's why all products, even waypoint which uses some different patterns than consul, nomad et. al, uses a single pkg scope for (nearly) all `cli.Command`s they define. I'll link below to Nomad's impl. below as a model.

The name of the `cmd` package differs between sets of products: consul, vault, and nomad vs waypoint. In the first 3 products it's called `command`, though waypoint differs and calls it `cli` while exposing a main function. I believe this is because they also host other UIs like a web server. Let's follow the most common example and use the pkg name `command`. Then we can use a `command.go` file to organize all of our commands like Nomad does, and split out files as needed. It's worth noting that Nomad does have a separate pkg in `command/agent` but that command is sufficiently complex to account for it.

Here's the Nomad layout that we can reapply: https://github.com/hashicorp/nomad/tree/main/command

agent/agent.go
```diff
@@ -179,11 +179,6 @@ func (a *Agent) Run() []error {
 		a.l.Error("Failed to cleanup after the run", "error", errCleanup)
 	}
 
-	a.l.Info("Writing summary of products and ops to standard output")
```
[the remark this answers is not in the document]
nwchandler: It was. I pulled it up into the enclosing command. It kind of felt like reporting on the status of the run was more of a command "thing" than an agent.Run "thing" to me.

main.go
```diff
@@ -1,275 +1,117 @@
...
-				"terraform", config.TFE,
-				"vault", config.Vault,
-			)
+func mainHelp(commands map[string]cli.CommandFactory) string {
+	// NOTE: This is a slightly modified copy of mitchellh/cli.BasicHelp so that we can include some general
+	// application usage information in it.
+	var buf bytes.Buffer
+	buf.WriteString(fmt.Sprintf("Usage: %s [--version] [--help] <command> [<args>]\n\n", appName))
+
+	appUsage := strings.TrimSpace(`
```
groovemonkey, line 69: This is a better usage string than 99% of Linux utilities; thank you.

groovemonkey (approved):
I've never worked with hcli, but this looks clean and navigable to me. Thanks.

PR #226, opened 2022-09-06
Add Do and DoSync runners

groovemonkey:
## Overview
Change the runner interface so Run returns a collection of ops.
    Products store the results of these ops under the id of the runner.

Add “do” runner that can accept a list of runners and calls .Run() on each in a goroutine.
    Return all results as []op.Op 
    We’ll need to change the runner return to []op.Op

Add “do-sync” runner that can accept a list of runners and calls .Run() for each in order.
    Check the Status of the result op and if it’s not Success, abort returning all existing ops.

## Notes
- runners don't currently log. that breaks logging -- to get around this, I'm passing the product logger (which is really the agent logger) into Do/DoSync.
- Host and nomad runners run inside of a `Do` runner (async): `return []runner.Runner{runner.NewDo(l, runners)}`

- I've written this intentionally so that Do and DoSync can be nested -- it ain't pretty, but it's fine:
```
// runners := []runner.Runner{} // whatever product runners you start with
syncRunners := []runner.Runner{
runner.NewCommander("uname -a", "string", cfg.Redactions),
runner.NewCommander("echo foobar", "string", cfg.Redactions),
}
// Execute asynchronously: Do inside of DoSync
allRunners := []runner.Runner{runner.NewDoSync(l, append(syncRunners, runner.NewDo(l, runners)))}
return allRunners, nil
```

- Do/DoSync runners could add their own "Op" result to the end of the op-slice they return from all their runners implemented in dosync.go). Is that how we want to do this?


product/product.go
```diff
@@ -110,12 +112,12 @@ func (p *Product) Filter() error {
 // CommanderHealthCheck employs the CLI to check if the client and then the agent are available.
 func CommanderHealthCheck(client, agent string) error {
 	checkClient := runner.NewCommander(client, "string", nil).Run()
-	if checkClient.Error != nil {
-		return fmt.Errorf("client not available, healthcheck=%v, result=%v, error=%v", client, checkClient.Result, checkClient.Error)
+	if checkClient[0].Error != nil {
```
groovemonkey, line 115: I'm thinking of maybe changing Op.Error (etc.) to Op.Errors() which makes the difference between multi-op runners (do/dosync groupings) and single-op runners (all others) transparent. That would save a lot of this ugly/brittle `op[0]` stuff.

runner/do.go
```diff
@@ -0,0 +1,70 @@
...
+	var mu sync.Mutex
+
+	// Start WaitGroup that will wait for all of our Runners
+	wg := new(sync.WaitGroup)
+	wg.Add(len(d.Runners))
+
+	for i, r := range d.Runners {
+		d.log.Info("running operation", "runner", r.ID())
+		go func(m *sync.Mutex, wg *sync.WaitGroup, sm *map[int][]op.Op, r Runner, ridx int) {
+			// Dereference pointer TODO(is there a better way?)
```
nwchandler, line 45: Since the map isn't needed until we have the lock, an alternative way (not sure it's better or worse, but it avoids a variable declaration) would be to do something like `(*sm)[ridx] = o` after the lock.

mkcp (approved):
Just realized I forgot to actually approve the branch, as discussed. Dave and I have working on this together and have iterated through two approaches for returning collections of results. Constraining Op.Result to a `map[string]any` allows us to build around nested trees of ops, and was easier to work with than []op.Op. do.Do and do.Sync leverage the results changes to return their result with multiple ops, and each previous runner has been changed to return results in a semantic key.

PR #228, opened 2022-09-12
commander: check commands before trying to run them

groovemonkey:
- create the cross-platform util.HostCommandExists() which checks to see if a command exists
- commander.Run() now uses this function to abort early (with an op.Skip, not an op.Fail)
- iptables runner now passes through commander status instead of wrapping all errors in op.Fail, which exercises HostCommandExists() on Docker containers that have no `iptables` command
  - bugfix: panic with nil op.Result in iptables runner

### DIFF
--- a/command/run.go
+++ b/command/run.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"os/exec"
 	"path"
 	"strings"
 	"time"
@@ -16,6 +15,7 @@ import (
 	"github.com/hashicorp/hcdiag/agent"
 	"github.com/hashicorp/hcdiag/hcl"
 	"github.com/hashicorp/hcdiag/product"
+	"github.com/hashicorp/hcdiag/util"
 )
 
 // seventyTwoHours represents the duration "72h" parsed in nanoseconds
@@ -195,7 +195,8 @@ func (c *RunCommand) Run(args []string) int {
 }
 
 // configureLogging takes a logger name, sets the default configuration, grabs the LOG_LEVEL from our ENV vars, and
-//  returns a configured and usable logger.
+//
+//	returns a configured and usable logger.
 func configureLogging(loggerName string) hclog.Logger {
 	// Create logger, set default and log level
 	appLogger := hclog.New(&hclog.LoggerOptions{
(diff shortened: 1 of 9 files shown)


groovemonkey:
Just tested this on an ageing Windows machine. 👍

runner/commander.go
```diff
@@ -33,13 +34,37 @@ func (c Commander) ID() string {
...
 
+// CommandExists is a cross-platform function that returns true if a command exists on the host
+func CommandExists(command string) bool {
+	var lookup string
+
+	// Strip command of args; we're only testing the first token
+	command = strings.Fields(command)[0]
+
+	// Set appropriate lookup command based on OS
+	if runtime.GOOS == "windows" {
```
groovemonkey, line 45: I just realized that I can implement this via exec.LookPath(), which might be cleaner and faster (and just as cross-platform as this). Happy to hear your thoughts.

mkcp: Let's use `exec.LookPath()` in the implementation and make some adjustments to the error returns. CommandExists ought to return a `CommandNotFoundError` while wrapping the error from `exec.LookPath()`.  `ShellExecError` has a good example of how to implement the Unwrap() method and include the wrapped message in the `Error()` message.

mkcp (approved):
👍 Good change. Added a small nitpick with the error impl. and handling and preemptively approving.

runner/commander.go
nwchandler: Could `command.autoDetectCommand` (in `command/run.go`) be extracted out to someplace shared and used for this? I think it does what is being looked for here, but it's just not exported currently. (`autoDetectCommand` could probably be given a more general name as well.)

groovemonkey: Thanks, both of you -- I've repurposed `command.autoDetectCommand` to fit Kit's description and moved to the `util` package (it just didn't seem right to put it into the command runner and then import that entire package in run.go). Let me know if there's a better place for it.

command/run.go
```diff
@@ -195,7 +195,8 @@ func (c *RunCommand) Run(args []string) int {
 }
 
 // configureLogging takes a logger name, sets the default configuration, grabs the LOG_LEVEL from our ENV vars, and
-//  returns a configured and usable logger.
+//
+//	returns a configured and usable logger.
```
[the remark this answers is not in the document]
groovemonkey: D'oh, I keep screwing these up. Thanks, will pin that template until I remember the categories.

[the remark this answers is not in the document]
groovemonkey: Woah, thanks for finding that. I blame an over-eager `go fmt` on save.

nwchandler (approved):

PR #232, opened 2022-09-14
Fix --includes by decoupling from global since/until Time

groovemonkey:
This fixes a sneaky little bug where users' `-includes` arguments were being ignored. It removes `CopyIncludes()`' dependency on global/agent config (it was passing global `since` and `until` time values into `NewCopier()`, which was leading to lots of ignored files by default).

After this change, users can just pass e.g. `-includes='/etc/consul.d/*'` or `-includes='/etc/consul.d'` and have the files they need for troubleshooting copied as expected.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -298,7 +298,8 @@ func (a *Agent) CopyIncludes() (err error) {
 		}
 
 		a.l.Debug("getting Copier", "path", f)
-		o := runner.NewCopier(f, dest, a.Config.Since, a.Config.Until, nil).Run()
+		// In this case, newCopier doesn't need real since/until; if I am passing the -includes CLI flag, I want matching files regardless of creation/mod timestamps
+		o := runner.NewCopier(f, dest, time.Time{}, time.Now(), nil).Run()
 		if o.Error != nil {
 			return o.Error
 		}
--- a/changelog/232.txt
+++ b/changelog/232.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+agent: decouple -includes from global since/until times, so it just copies files without users having to worry about modification time
+```


agent/agent.go
```diff
@@ -298,7 +298,8 @@ func (a *Agent) CopyIncludes() (err error) {
 		}
 
 		a.l.Debug("getting Copier", "path", f)
-		o := runner.NewCopier(f, dest, a.Config.Since, a.Config.Until, nil).Run()
+		// In this case, newCopier doesn't need real since/until; if I am passing the -includes CLI flag, I want matching files regardless of creation/mod timestamps
+		o := runner.NewCopier(f, dest, time.Time{}, time.Now(), nil).Run()
```
nwchandler, line 302: This seems like it's probably reasonable to me, but I'm not actually sure if the existing/previous behavior was by design. Perhaps @mkcp may know?

changelog/232.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
```
nwchandler, line 1: Sorry to be the changelog pedant, but if it turns out that this was a bug (pending Kit's response to the comment above), we should make the release note type `bug`. That will keep it grouped in the right place when the changelog is generated. 🙂

agent/agent.go
mkcp: The previous behavior was by design. `-includes` is in a bit of an awkward spot because it was the first way we had to copy files and predates both Copy runners and HCL. You can see some of the design sprawl in both `-include-since` becoming an alias for `-since` after we converged on a shared time range for all hcdiag runs. Combining a wildcard with an `-include-since` is how we filtered out log files into a time range. 

The shared time range for all runners is important for avoiding concurrency bugs because `time.Now()` will have different values depending on when it's called. We want one value for now shared between every runner, so that now can be understood. Effectively, "now" is the state of the system when a user presses enter to start a run.

Looking at this some more, I'm not sure that `-includes` has a clear niche anymore. Changing it to mean "always copy this file" rather than "copy this file if it's within hcdiag's time range" makes sense to give it a new niche. Although, I'm more inclined to deprecate it so users are always working directly with a `Copy` runner rather than what became a flagged alias for one. We've kept it around as a convenient shorthand, but if we're struggling with how the semantics should be, then I imagine it's even more confusing for end users.

nwchandler: Thanks for the background, Kit! Given all of that, I would favor deprecating it as well. It seems like `Copy` provides what `-includes` does (and more), and it would be a more consistent user experience to use `Copy` for all file-copy needs.

changelog/232.txt
groovemonkey: Yeah, I don't know why I decided to go crazy with these -- that's the last time, I promise! I've stickied all the valid changelog release-note types. Sorry about that.

agent/agent.go
groovemonkey: Thanks -- I appreciate the context; happy to close this PR.

groovemonkey:
Closing because this behavior is intentional -- we may just deprecate -includes entirely, in favor of explicit use of Copy runners.

PR #233, opened 2022-09-19
Doc improvements: custom config examples, bugfix, FAQ addition

groovemonkey:
This PR adds:

- a spelling fix with one example of `-includes`
- custom config examples from the last few weeks of working with users
- a FAQ addition that points out that hcdiag is a single-host binary, and does not collect information from more than one host at a time.

### DIFF
--- a/changelog/233.txt
+++ b/changelog/233.txt
@@ -1,3 +1,3 @@
 ```release-note:docs
-docs: add custom config examples, faq content"
+docs: add custom config examples, faq content
 ```
--- a/changelog/233.txt
+++ b/changelog/233.txt
@@ -1,3 +1,3 @@
-```release-note:docs
+```release-note:improvement
 docs: add custom config examples, faq content
 ```
--- a/changelog/233.txt
+++ b/changelog/233.txt
@@ -1,0 +1,3 @@
+```release-note:docs
+docs: add custom config examples, faq content"
+```
--- a/docs/custom-config.md
+++ b/docs/custom-config.md
@@ -69,3 +69,77 @@ product "consul" {
 Beginning with version `0.4.0`, `hcdiag` supports redactions. Redactions enable users to tell `hcdiag` about patterns of text that should be omitted from the results bundle.
 
 Read more about them [here](./redactions.md).
+
+## Custom Config Examples
+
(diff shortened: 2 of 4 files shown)


changelog/233.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:docs
+docs: add custom config examples, faq content"
```
nwchandler, line 2: Looks like there's an extra `"` at the end
```suggestion
docs: add custom config examples, faq content
```

docs/custom-config.md
```diff
@@ -69,3 +69,77 @@ product "consul" {
...
+## Custom Config Examples
+
+### Collect additional Consul information
+
+**Example:** we want a `-consul` run to collect additional information from several commands:
+
+* The Consul memberlist: `consul members`
+* Raft peer information: `consul operator raft list-peers`
+* A custom journalctl command that's different from the builtin: `journalctl -u consul --since=yesterday --output=json`
+* another custom command (in this case, just running `cat` to collect a file's contents into the bundle `Results.json` file with no redactions).
```
nwchandler, line 82: Since the previous three bullet points start with capital letters and don't include ending punctuation, I think it would be worth doing the same here.
```suggestion
* Another custom command (in this case, just running `cat` to collect a file's contents into the bundle `Results.json` file with no redactions)
```

changelog/233.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:docs
```
nwchandler, line 1: There isn't a section in the template for a note type of `docs` so this would not show up in the changelog. We could add it to the template, or else we could make this a note type that will be in the changelog, such as `improvement`. Rather than changing the template, though, I think just calling it an improvement would be best.

```suggestion
```release-note:improvement
```

nwchandler (changes requested):
I left a couple minor suggestions on the text formatting and one on the changelog note type that needs to be updated to ensure this makes it into the release notes. Otherwise, it looks good to me.

groovemonkey:
Thanks for catching that! I've committed your suggestions.

nwchandler (approved):
LGTM

PR #234, opened 2022-09-22
start gathering more basic system info for TFE

groovemonkey:


mkcp (commented):
Just wanted to give you a heads up that there's a lot of duplication here in commands like `date, ip addr, fstab, hosts` that are already being collected as runners in the host pkg.

We implemented as many of the replicated command as we had time for as host runners so they could be offered in a product-agnostic way. For the commands that don't have host runners, there is likely an issue in the backlog to add it to `host`, or there's an existing host runner that collects the same data from a different command.

We don't necessarily need specific runner types for all of these, but one of hcdiag's main value propositions is cross-product consistency.

groovemonkey:
That makes sense -- this was just a first crack at turning the support bundle analysis into hcdiag code before I forgot what I was doing. I'll move most of this into the host runners and dedupe what's already collected there.

groovemonkey:
I'm going to close this for now -- I'll use it as a reference when I have time to jump back onto this project!

PR #235, opened 2022-09-27
Always Clean up Temporary Directory

nwchandler:
This addresses a bug, in which the temporary directory created by the agent would only be cleaned up in the event that hcdiag ran to its expected completion. Certain error cases, such as product healthcheck failures, would end up leaving the temporary directory on the filesystem. We now should clean up after ourselves, regardless of whether the agent run completes successfully or whether it errors.


mkcp:
LGTM 👍

mkcp (approved):

PR #236, opened 2022-09-28
Check for Container Existence Before Getting Logs

nwchandler:
This merge adds a check before looking for Docker container logs to ensure the requested container actually exists. Previously, when a user ran hcdiag on a machine that had docker installed, but there were no matching containers, we created empty log files in the results bundles rather than skipping the log check.

### DIFF
--- a/runner/log/docker.go
+++ b/runner/log/docker.go
@@ -85,10 +85,7 @@ func (d Docker) containerExists() bool {
 	// attempt to inspect the container by name, to ensure it exists
 	cmd := fmt.Sprintf("docker container inspect %s > /dev/null 2>&1", d.Container)
 	o := runner.NewSheller(cmd, d.Redactions).Run()
-	if o.Error != nil {
-		return false
-	}
-	return true
+	return o.Error == nil
 }
 
 var _ error = DockerNotFoundError{}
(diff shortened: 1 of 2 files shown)


runner/log/docker.go
```diff
@@ -49,18 +49,20 @@ func (d Docker) Run() op.Op {
...
 	// Retrieve logs
 	cmd := DockerLogCmd(d.Container, d.DestDir, d.Since)
 	o = runner.NewSheller(cmd, d.Redactions).Run()
-	// NOTE(mkcp): If the container does not exist, docker will exit non-zero and it'll surface as a ShellExecError.
-	//  The result actionably states that the container wasn't found. In the future we may want to scrub the result
-	//  and only return an actionable error message
 	if o.Error != nil {
 		return op.New(d.ID(), o.Result, o.Status, o.Error, runner.Params(d))
 	}
-	if o.Result == "" {
```
nwchandler, line 61: Because we redirect the shell output, it appeared in testing that the result was always `""`. This meant that we could gather logs but still warn that there were no logs. I removed this check and warning as a result of this behavior, so we wouldn't end up with conflicting information.

mkcp (approved):
This all makes sense - +1 for more robust error handling. 👍

PR #237, opened 2022-09-29
add deprecation warnings for the -includes flag

groovemonkey:
We're going to deprecate `-includes` in favor of using copy runners in future versions of hcdiag. This change adds deprecation warnings in appropriate places:

1. When using the `-includes` flag:

```
# ./hcdiagincludes -includes=/etc/lsb-release

2022-09-29T12:47:21.706Z [WARN]  hcdiag: DEPRECATION NOTICE: The '-includes' option will be removed in an upcoming version of hcdiag. Please use HCL copy blocks instead.
```

2. When viewing help/usage text, i.e. `hcdiag run --help`:

```
  -includes
     DEPRECATED: Files or directories to include (comma-separated,
     file-*-globbing available if 'wrapped-*-in-single-quotes'); e.g.
     '/var/log/consul-*,/var/log/nomad-*'. NOTE: This option will be
     removed in an upcoming version of hcdiag. Please use HCL copy
     blocks instead
```

### DIFF
--- a/changelog/237.txt
+++ b/changelog/237.txt
@@ -1,0 +1,3 @@
+```release-note:deprecation
+deprecation: -includes flag will be removed in a future version of hcdiag
+```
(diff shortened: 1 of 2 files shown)


nwchandler (approved):
LGTM

PR #238, opened 2022-10-03
Track Runner Durations

groovemonkey:
Adds start and end times to all runners, and writes each op's duration (in ns) to manifest.

I originally thought about doing something fancy from outside the runners to save lines of code, but I think this ended up cleaner.

Ops now have a StartTime and EndTime, and all the runners have the same change at each possible return:

```
startTime := time.Now()
...
return Op.New(..., startTime, time.Now())
```

The duration for each Op (currently in nanoseconds, because host runners are really fast) is added to Manifest.json.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -246,7 +246,8 @@ func (a *Agent) TempDir() string {
 }
 
 // CreateTemp Creates a temporary directory so that we may gather results and files before compressing the final
-//  artifact.
+//
+//	artifact.
 func (a *Agent) CreateTemp() error {
 	tmp, err := os.MkdirTemp(a.Config.Destination, a.TempDir())
 	if err != nil {
@@ -311,7 +312,8 @@ func (a *Agent) CopyIncludes() (err error) {
 
 // RunProducts executes all ops for this run.
 // TODO(mkcp): We can avoid locking and waiting on results if all results are generated async. Then they can get streamed
-//  back to the dispatcher and merged into either a sync.Map or a purpose-built results map with insert(), read(), and merge().
+//
+//	back to the dispatcher and merged into either a sync.Map or a purpose-built results map with insert(), read(), and merge().
 func (a *Agent) RunProducts() error {
 	// Set up our waitgroup to make sure we don't proceed until all products execute.
 	wg := sync.WaitGroup{}
@@ -356,10 +358,14 @@ func (a *Agent) RunProducts() error {
 func (a *Agent) RecordManifest() {
 	for name, ops := range a.results {
 		for _, o := range ops {
+			// duration string, in nanoseconds
+			dur := fmt.Sprintf("%d nanoseconds", o.EndTime.Sub(o.StartTime).Nanoseconds())
+
 			m := ManifestOp{
-				ID:     o.Identifier,
-				Error:  o.ErrString,
-				Status: o.Status,
+				ID:       o.Identifier,
+				Error:    o.ErrString,
+				Status:   o.Status,
+				Duration: dur,
 			}
 			a.ManifestOps[string(name)] = append(a.ManifestOps[string(name)], m)
 		}
--- a/agent/manifest.go
+++ b/agent/manifest.go
@@ -7,7 +7,8 @@ import (
 // ManifestOp provides a subset of op state, specifically excluding results, so we can safely render metadata
 // about ops without exposing customer info in manifest.json
 type ManifestOp struct {
-	ID     string    `json:"op"`
-	Error  string    `json:"error"`
-	Status op.Status `json:"status"`
+	ID       string    `json:"op"`
+	Error    string    `json:"error"`
+	Status   op.Status `json:"status"`
+	Duration string    `json:"duration"`
 }
--- a/op/op.go
+++ b/op/op.go
@@ -2,6 +2,7 @@ package op
 
 import (
 	"fmt"
+	"time"
 )
 
 // Status describes the result of an operation.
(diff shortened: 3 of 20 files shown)


mkcp (changes requested):
Good change overall - having timestamps in Results.json, and durations in Manifest.json readouts is a good boundary to have.

A few changes requested:
1: On op, the new fields StartTime and EndTime have the type in the name which isn't idiomatic. This isn't necessary, we can use Start and End. 

2: This is a big scope increase, so resolving in a follow up is valid.  `op.New`'s params are  getting sprawly. There's a few ways we can address this. First, we ought to add another New fn that can fill in the value for End: maybe `op.NewEndNow(id, result, status, err, params, startAt)` or `op.NewWithEnd`.

Another way to handle init complexity is with a config struct, though if we didn't have to store err.Error() on `ErrString` we could use the zeroed value of `op.Op` itself rather than op.New forgoing a need to mirror the struct. Then any helpers could merge in batches of params onto any `op.Op` needed. IIRC ErrString is there to help grab the err.Error() - this is simple to fix in our json marshaling though.

Not sure what the best path forward is here - what do you think?

nwchandler:
> Good change overall - having timestamps in Results.json, and durations in Manifest.json readouts is a good boundary to have.
> 
> A few changes requested: 1: On op, the new fields StartTime and EndTime have the type in the name which isn't idiomatic. This isn't necessary, we can use Start and End.
> 
> 2: This is a big scope increase, so resolving in a follow up is valid. `op.New`'s params are getting sprawly. There's a few ways we can address this. First, we ought to add another New fn that can fill in the value for End: maybe `op.NewEndNow(id, result, status, err, params, startAt)` or `op.NewWithEnd`.
> 
> Another way to handle init complexity is with a config struct, though if we didn't have to store err.Error() on `ErrString` we could use the zeroed value of `op.Op` itself rather than op.New forgoing a need to mirror the struct. Then any helpers could merge in batches of params onto any `op.Op` needed. IIRC ErrString is there to help grab the err.Error() - this is simple to fix in our json marshaling though.
> 
> Not sure what the best path forward is here - what do you think?

With regards to point number 2, I wonder if the functional options pattern would be useful to us: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

groovemonkey:
> 2: This is a big scope increase, so resolving in a follow up is valid. `op.New`'s params are getting sprawly. There's a few ways we can address this. First, we ought to add another New fn that can fill in the value for End: maybe `op.NewEndNow(id, result, status, err, params, startAt)` or `op.NewWithEnd`.
> 
> Another way to handle init complexity is with a config struct, though if we didn't have to store err.Error() on `ErrString` we could use the zeroed value of `op.Op` itself rather than op.New forgoing a need to mirror the struct. Then any helpers could merge in batches of params onto any `op.Op` needed. IIRC ErrString is there to help grab the err.Error() - this is simple to fix in our json marshaling though.
> 
> Not sure what the best path forward is here - what do you think?

Thanks for pointing that out -- let's discuss which approach to take at our next quick hcli-triumverate meeting; I'll break it out into a separate ticket in the meantime!

agent/agent.go
```diff
@@ -246,7 +246,7 @@ func (a *Agent) TempDir() string {
 }
 
 // CreateTemp Creates a temporary directory so that we may gather results and files before compressing the final
-//  artifact.
+// artifact.
```
groovemonkey, line 249: Removing these spaces prevents `go fmt` from going wild and adding an empty comment line when I save -- hope that's ok!

nwchandler: Yeah, where there are extra spaces in doc comments, we should remove them. Go Doc interprets these as pre-formatted text (code blocks). See the comment for agent.CreateTemp() at https://pkg.go.dev/github.com/hashicorp/hcdiag@v0.4.0/agent#Agent.CreateTemp
image

mkcp: As a frequent space-placer, I'd been adding the space because it makes goland highlight the next line with the same comment font. Didn't realize it was breaking things downstream. No need to keep the spaces!

mkcp:
> > Good change overall - having timestamps in Results.json, and durations in Manifest.json readouts is a good boundary to have.
> > A few changes requested: 1: On op, the new fields StartTime and EndTime have the type in the name which isn't idiomatic. This isn't necessary, we can use Start and End.
> > 2: This is a big scope increase, so resolving in a follow up is valid. `op.New`'s params are getting sprawly. There's a few ways we can address this. First, we ought to add another New fn that can fill in the value for End: maybe `op.NewEndNow(id, result, status, err, params, startAt)` or `op.NewWithEnd`.
> > Another way to handle init complexity is with a config struct, though if we didn't have to store err.Error() on `ErrString` we could use the zeroed value of `op.Op` itself rather than op.New forgoing a need to mirror the struct. Then any helpers could merge in batches of params onto any `op.Op` needed. IIRC ErrString is there to help grab the err.Error() - this is simple to fix in our json marshaling though.
> > Not sure what the best path forward is here - what do you think?
> 
> With regards to point number 2, I wonder if the functional options pattern would be useful to us: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

@nwchandler This most applies when you need to optionally compose init behaviors. Ideally we'll be able to factor out `op.ErrString` and we can avoid initializers all together!

mkcp (approved):
LGTM 👍

nwchandler:
> @nwchandler This most applies when you need to optionally compose init behaviors. Ideally we'll be able to factor out `op.ErrString` and we can avoid initializers all together!

@mkcp I forgot to reply (to myself) earlier today, but as I thought about this some more, it also seems like the options pattern is probably not necessary for primarily data structs. As you suggested, altering behavior would be a better reason to implement options. So, yeah, I'm all for factoring out `op.ErrString` and getting rid of the initializer all together! 😄

groovemonkey:
OK, I've done the merge, attempting to be careful while I do it. Tests are passing. If someone could take a quick look at `iptables` and `journald` runners to make sure I kept the right code, I'd appreciate it <3.

agent/agent.go
groovemonkey: Just created https://github.com/hashicorp/hcdiag/pull/243 for this

PR #239, opened 2022-10-03
Improve Handling of Commander Args with Spaces

nwchandler:
This intends to improve hcdiag's ability to handle commander args that include spaces in them on Linux/Unix OS's. Previously, such arguments would be inappropriately split, even if they were wrapped within single quotes, which caused the command to be unable to make sense of the arguments. A library has been included which should alleviate this issue, and tests were added around the capability.

Note that no changes are made to Windows command argument processing. The underlying mechanism for Windows command calls is done differently than *nix, so the existing behavior worked for Windows.


mkcp (approved):
This is good! Small revision requested for parseCommand's error handling but otherwise 👍

runner/commander.go
```diff
@@ -96,6 +99,70 @@ func (c Commander) Run() op.Op {
 	}
 }
 
+type parsedCommand struct {
+	cmd  string
+	args []string
+	err  error
+}
+
+func parseCommand(command string) parsedCommand {
```
mkcp, line 108:
```suggestion
func parseCommand(command string) (parsedCommand, error) {
```

Make sure to mutlireturn the error as a part of the fn signature to keep the api idiomatic. You always want to return the error itself for immediate handling. It's assumed that the return value isn't usable if err != nil, and this could create a gotcha in the future if the return value is assumed to always be ok by a consumer.

nwchandler: Good call!

PR #241, opened 2022-10-04
Add Bundle Location to End-of-Run Summary

nwchandler:
This merge adds a message indicating the run completed, along with a notification of where the bundle was written, to the end of the diagnostic run. The hope is this will make it easier for users to locate the bundle.

Along with this, the merge moves the responsibility of displaying the summary from the agent to the command package. This was intended to maintain the CLI as the intermediary between the user and the agent.


agent/agent.go
```diff
@@ -507,83 +505,12 @@ func (a *Agent) Setup() error {
...
-	}
-
-	format += "\n"
-
-	return fmt.Sprintf(format, strValues...)
+// ResultsDest is provided for read-only access to the destination where the agent writes its results. The particular
+// destination is determined by the agent while running; it is unexported to avoid accidental overwrite by external
+// packages. However, its value is useful to know for downstream user interaction, so this method intends to provide
+// that ability.
+func (a *Agent) ResultsDest() string {
```
nwchandler, line 512: I waffled on just exporting the destination field or adding a public getter method. The doc comment describes why I landed on this, but I would be open to other opinions here for sure.

mkcp (approved):
Good change - helps reinforce the boundary between CLI behavior and the agent. 👍

PR #242, opened 2022-10-05
add a generator that only handles strings and command/shell

mkcp:
Just messing around with generators before we have the machinery to pass result values around.


mkcp:
Closing this PR out as part of cleanup.

PR #243, opened 2022-10-06
fix extra spaces in go comments

groovemonkey:
These spaces mess with Go's automatic documentation formatting, so I've removed them.


nwchandler (approved):
LGTM

PR #245, opened 2022-10-07
Count nested operations in StatusCounts

groovemonkey:
This change adds op.WalkStatuses(), which recursively walks a tree of result ops to get statuses.

### DIFF
--- a/op/op.go
+++ b/op/op.go
@@ -59,7 +59,8 @@ func StatusCounts(ops map[string]Op) (map[Status]int, error) {
 	}
 
 	// Create our accumulator and Walk ops
-	acc := make(map[Status]int, 0)
+	// We'll always have at least len(ops) Ops; more if there are nested Ops
+	acc := make(map[Status]int, len(ops))
 	statuses := WalkStatuses(acc, m)
 
 	if val, ok := statuses[""]; ok {
--- a/op/op.go
+++ b/op/op.go
@@ -58,10 +58,7 @@ func StatusCounts(ops map[string]Op) (map[Status]int, error) {
 		m[k] = v
 	}
 
-	// Create our accumulator and Walk ops
-	// We'll always have at least len(ops) Ops; more if there are nested Ops
-	acc := make(map[Status]int, len(ops))
-	statuses := WalkStatuses(acc, m)
+	statuses := WalkStatuses(m)
 
 	if val, ok := statuses[""]; ok {
 		return nil, fmt.Errorf("op.StatusCounts received ops that did not have a status, count=%d", val)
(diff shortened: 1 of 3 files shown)


op/op.go
```diff
@@ -52,12 +52,39 @@ func New(id string, result map[string]any, status Status, err error, params map[
...
-		}
-		statuses[o.Status]++
+	// copy our input into a new map that conforms to our Walk input type
+	m := make(map[string]any, len(ops))
+	for k, v := range ops {
+		m[k] = v
+	}
+
+	// Create our accumulator and Walk ops
+	acc := make(map[Status]int, 0)
```
nwchandler, line 62: I'm unclear on why this map would be initialized with a size of 0. 🤔 Would it be better to omit the size or even to declare `var acc map[Status]int` instead?

op/op.go
```diff
@@ -52,12 +52,39 @@ func New(id string, result map[string]any, status Status, err error, params map[
...
+	statuses := WalkStatuses(acc, m)
+
+	if val, ok := statuses[""]; ok {
+		return nil, fmt.Errorf("op.StatusCounts received ops that did not have a status, count=%d", val)
 	}
 	return statuses, nil
 }
+
+// WalkStatuses performs a depth-first search of a tree of results and accumulates a map of status counts
+func WalkStatuses(statuses map[Status]int, results map[string]any) map[Status]int {
```
nwchandler, line 72: We're mutating `statuses` here, and then returning it at the end. Perhaps we could avoid the mutation, though. What if `WalkStatuses` took only a `results` map, and it had an inner function defined, which would take a status map. We could recurse on the inner function and then return what it built up at the end of `WalkStatuses`. This would, I think, both make the calling of the `WalkStatuses` easier (no need to create an accumulator) and also avoid accidental mutation.

nwchandler: `walkStatuses` could also just be an unexported function as well - the main idea is, can we avoid mutation in the public API?

op/op.go
groovemonkey: Totally right; thanks for catching that!

op/op.go
mkcp: Not certain it's the ideal implementation, but it does let us pass in partially constructed accumulators for testing or for more elaborate concurrent reducers. You could reasonably do the same thing without passing in an accumulator though, treating it as the steps: `split/partition results` -> `map over result subset and return map[Status]int` -> `merge all map[Status]int`. Also, The accumulator is passede by so we don't have to worry about changes from outside goroutines.

mkcp: I was mistaken! maps are like channels not slices and get passed by reference even without an explicit pointer!

op/op.go
groovemonkey: I changed this to initialize to len(ops), which is the minimum length this will have (if there are no nested ops).

op/op.go
groovemonkey: Thanks -- used @nwchandler 's inner-closure version of this function. Let me know if that's good to go!

nwchandler (approved):
Looks good to me!

PR #248, opened 2022-10-11
add runners for 'consul operator raft list-peers' and /v1/agent/members

groovemonkey:
Both runners use the local Consul agent's cache to minimize impact on the cluster.

`consul operator raft list-peers` gives more information than just the two HTTP queries we already have (`/status/{peers,leader}`), so I’m adding it via a Commander runner.


mkcp (approved):
LGTM 👍 ✨

PR #252, opened 2022-10-13
Skip Displaying Run Summary on Dry Runs

nwchandler:
This merge addresses a UX bug introduced in #222 . With the changes introduced there, we began displaying a summary report at the end of dry runs, however the output was nonsensical because no runners had executed, and no results were written to disk. This change puts some condition checking in place to ensure we do not write the output summary when a user does a dry run.


mkcp (approved):
LGTM! 👍

PR #255, opened 2022-10-19
Make Default Runners Concurrent

nwchandler:
This merge changes the default runners within products to execute concurrently rather than serially. Specifically, the default runners are now wrapped within a Do runner, which handles concurrent execution. One small exception is in TFE, where one runner generates a support bundle, and another one copies the bundle; these have been further wrapped within a DoSync block inside of the wrapping Do block.

Additionally, the --serial flag is now deprecated. The behavior of the serial flag is unclear when runners execute concurrently by default. The same behavior may be configured using do-sync blocks within HCL if needed.


groovemonkey (approved):
I love it when a plan comes together. This looks good!

PR #257, opened 2022-10-25
Rename runners to drop "er" endings

groovemonkey:
Individual runner struct names currently end with “er,” but we prefer them as simple Nouns instead of "`thing`-ers. “Commander” becomes “Command,” “Sheller” becomes “Shell,” etc.


hcl/hcl.go
```diff
@@ -184,20 +184,20 @@ func BuildRunners[T Blocks](config T, tmpDir string, c *client.APIClient, since,
...
 		if err != nil {
 			return nil, err
 		}
 		runners = append(runners, gets...)
 
 		// Identical code between Product and Host, but cfg's type must be resolved via the switch to access the fields
-		// Build copiers
-		copiers, err := mapCopies(cfg.Copies, redactions, dest)
+		// Build Copy runners
+		copys, err := mapCopies(cfg.Copies, redactions, dest)
```
nwchandler, line 196: Hmm... Should this be `copies`?

nwchandler (approved):
I left a comment about a potential variable spelling change for consideration, but otherwise, this looks good to me.

hcl/hcl.go
groovemonkey: Yeah, I went back and forth on this one. I can change it back.

nwchandler: Yeah, personally, I think it reads a little better - especially, since on this same line, we also have `mapCopies` and `cfg.Copies` both spelled that way.

groovemonkey: Never ask an Austrian to fix the grammar... 🇦🇹

PR #265, opened 2022-11-01
Add VaultDebug Runner

groovemonkey:
`vault.go` is fully configurable, and exposes all of `vault debug`'s command-line options.

It's also configurable via custom HCL config:

```hcl
// produces the following vault debug command:
// vault debug -compress=true -duration=30s -interval=0m10s -logformat=standard -metricsinterval=10s -output=/vault-hcdiag/hcdiag-2022-11-03T211126Z2282457017/VaultDebug.tar.gz -target=metrics
product "vault" {
  vault-debug {
    compress = true
    interval = "0m10s"
    duration = "30s"
    targets = ["metrics"]
  }
}
```

Highly configurable ConsulDebug and NomadDebug runners are planned next.


hcl/hcl.go
```diff
@@ -173,12 +187,12 @@ func Parse(path string) (HCL, error) {
 // BuildRunners steps through the HCLConfig structs and maps each runner config type to the corresponding New<Runner> function.
 // All custom runners are reduced into a linear slice of runners and served back up to the product.
 // No runners are returned if any config is invalid.
-func BuildRunners[T Blocks](config T, tmpDir string, c *client.APIClient, since, until time.Time, redactions []*redact.Redact) ([]runner.Runner, error) {
-	return BuildRunnersWithContext(context.Background(), config, tmpDir, c, since, until, redactions)
+func BuildRunners[T Blocks](config T, tmpDir string, debugDuration time.Duration, debugInterval time.Duration, c *client.APIClient, since, until time.Time, redactions []*redact.Redact) ([]runner.Runner, error) {
```
groovemonkey, line 190: These new runners need to use `debugDuration` and `debugInterval` as defaults. These values are usually passed down via `product.Config` but need to be individual params here to prevent `hcl` and `product` from mutually importing each other.

product/vault.go
```diff
@@ -82,7 +82,18 @@ func vaultRunners(ctx context.Context, cfg Config, api *client.APIClient, l hclo
 		runner.NewCommand("vault read sys/health -format=json", "json", cfg.Redactions),
 		runner.NewCommand("vault read sys/seal-status -format=json", "json", cfg.Redactions),
 		runner.NewCommand("vault read sys/host-info -format=json", "json", cfg.Redactions),
-		runner.NewCommand(fmt.Sprintf("vault debug -output=%s/VaultDebug.tar.gz -duration=%s -interval=%s", cfg.TmpDir, cfg.DebugDuration, cfg.DebugInterval), "string", cfg.Redactions),
+
+		// Cross-product SimpleDebug vs. Fully-Configurable VaultDebug
```
groovemonkey, line 86: Just a little demo of the new stuff.

hcl/hcl.go
mkcp: Once we have specific runners for debug we ought to deprecate the debugDuration and debugInterval flags - their main usage was to pave over us not having an hcl-integrated runner.

changelog/265.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:improvement
+runner: add debug runners to wrap our products' debug commands: SimpleDebug, NomadDebug, VaultDebug, and ConsulDebug.
```
mkcp, line 2: The changelog note mention Nomad and Consul which I think are cut for scope? We should update this release note to mention the specific additions in the PR.

runner/debug/simple.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/hashicorp/hcdiag/op"
+	"github.com/hashicorp/hcdiag/redact"
+	"github.com/hashicorp/hcdiag/runner"
+)
+
+var _ runner.Runner = SimpleDebug{}
+
+// The "Simple" debug wrapper can be used for Nomad, Vault, and Consul.
+// It only deals with flags common to all three 'debug' commands
+type SimpleDebug struct {
```
mkcp, line 16: Do we use or intend to use SimpleDebug for Nomad and Consul, or is the intent to give them their own debug runners like Vault? If it's the former than we should call it `Debug` or `ProductDebug` and if it's the latter then we shouldn't merge it in.

mkcp (commented):
Heya, good stuff overall! There's a lot of params needed to integrate with the debug commands but it looks like there's a solid approach taken. Before we merge into main, I have a few questions and requests for revision.

changelog/265.txt
groovemonkey: Ahh thanks, I knew that might bee too ambitious of a changelog entry to write :-)

runner/debug/simple.go
groovemonkey: Good point -- I've created tasks/stories for NomadDebug and VaultDebug, and was on the fence about keeping SimpleDebug. I can't think of a great reason to keep it, either, if you think the current (VaultDebug) approach works well.

runner/debug/debug.go
```diff
@@ -0,0 +1,75 @@
...
+		"Job":        true,
+		"Allocation": true,
+		"Deployment": true,
+		"Evaluation": true,
+		"Node":       true,
+		"Service":    true,
+	}
+
+	vaultOptFlag := "target"
+	vaultFilters := map[string]bool{
```
nwchandler, line 30: Do we know how often these change across products? I worry that we could end up tightly coupled to product versions.

runner/debug/vault.go
```diff
@@ -0,0 +1,123 @@
...
+	Redactions      []*redact.Redact
+}
+
+// VaultDebug represents a VaultDebug runner
+type VaultDebug struct {
+	Compress        string           `json:"compress"`
+	Duration        string           `json:"duration"`
+	Interval        string           `json:"interval"`
+	LogFormat       string           `json:"logformat"`
+	MetricsInterval string           `json:"metricsinterval"`
```
nwchandler, line 31: For consistency with other runners, I think snake_case may be more appropriate for the JSON tags:
```suggestion
	LogFormat       string           `json:"log_format"`
	MetricsInterval string           `json:"metrics_interval"`
```

runner/debug/vault.go
```diff
@@ -0,0 +1,123 @@
...
+	output string
+}
+
+func (d VaultDebug) ID() string {
+	return "VaultDebug"
+}
+
+func NewVaultDebug(cfg VaultDebugConfig, tmpDir string, debugDuration time.Duration, debugInterval time.Duration) *VaultDebug {
+	dbg := VaultDebug{
+		// No compression because the hcdiag bundle will get compressed anyway
```
nwchandler, line 44: I think the product debugs get compressed today, don't they? I would probably rather either keep it consistent or else be sure to capture this change in the changelog so that users have a way of expecting the change in behavior without reading the code.

groovemonkey: Yep, I originally planned to have the SimpleDebug runner be a drop-in replacement, with the more configurable runners mirroring product debug defaults. That's getting removed, and I'll change the default to `-compress=true`.

runner/debug/vault.go
groovemonkey: Great catch -- done!

runner/debug/debug.go
groovemonkey: I'm not sure, but I don't think it's frequent. Is it OK if I investigate this as part of the next two runners I build (Nomad and Consul) -- In general I'm happy to make this a simple passthrough and not try to track/enforce valid filter options for the underlying debug command.

hcl/hcl.go
groovemonkey: I'll create a ticket for that!

runner/debug/debug.go
nwchandler: Yeah, I think it's fine to investigate in the upcoming work items.

runner/debug/debug.go
```diff
@@ -0,0 +1,75 @@
...
+		return "", fmt.Errorf("invalid product used in debug.productFilterString(): %s", product)
+	}
+
+	for _, f := range filters {
+		if legalFilters[f] {
+			// includes leading space
+			filterString = fmt.Sprintf("%s -%s=%s", filterString, optFlag, f)
+		} else {
+			return "", fmt.Errorf("invalid filter string (%s) for %s used in debug.productFilterString()", f, product)
+		}
```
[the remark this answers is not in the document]
groovemonkey: Fixed

runner/debug/debug_test.go
```diff
@@ -0,0 +1,109 @@
...
+			cfg: VaultDebugConfig{
+				Compress:        "false",
+				LogFormat:       "standard",
+				MetricsInterval: "10s",
+				Targets:         []string{},
+				Redactions:      []*redact.Redact{},
+			},
+			productDuration: 5 * time.Minute,
+			productInterval: 45 * time.Second,
+			filterString:    "",
```
nwchandler, line 32: I think these `filterString` fields can be removed when they have empty strings, since that's the default for a field of type `string`.

runner/debug/debug.go
```diff
@@ -0,0 +1,12 @@
+package debug
+
+import "fmt"
+
+// filterArgs returns a string that contains one '-flagname=filter' pair for each filters element
+func filterArgs(flagname string, filters []string) string {
+	var arguments string
```
nwchandler, line 7: Since we're building up a string value, I think a `strings.Builder` might be a more intuitive approach than the repeated reassignment of "arguments + something else" to "arguments."

hcl/hcl.go
```diff
@@ -109,6 +112,7 @@ type DoSync struct {
 	Copies       []Copy        `hcl:"copy,block" json:"copies,omitempty"`
 	DockerLogs   []DockerLog   `hcl:"docker-log,block" json:"docker_log,omitempty"`
 	JournaldLogs []JournaldLog `hcl:"journald-log,block" json:"journald_log,omitempty"`
+	VaultDebugs  []VaultDebug  `hcl:"vaultdebug,block" json:"vaultdebug,omitempty"`
```
nwchandler, line 115: Looks like the hcl should be `vault-debug` and json should be `vault_debug` for consistency with the other runners.

runner/debug/debug.go
nwchandler: I updated this while Dave is away.

hcl/hcl.go
nwchandler: I updated this while Dave is away.

mkcp (approved):
LGTM

PR #269, opened 2022-11-10
Migrate Command Runner Constructors to Use Config Objects

nwchandler:
This merge updates the Command constructors to leverage a new CommandConfig object, rather than parameters. The intention is to simplify future feature additions by limiting changes to the function signatures of these constructors.


runner/host/iptables.go
```diff
@@ -38,12 +38,21 @@ func (r IPTables) ID() string {
 
 func (r IPTables) Run() op.Op {
 	startTime := time.Now()
-	if r.OS == "linux" {
+	if r.OS != "linux" {
```
nwchandler, line 41: Unrelated to the change at hand, but this looked like a little bug that had crept in as part of a recent merge.

groovemonkey (approved):
Yeah, this could work -- it's certainly a nicer interface for people who aren't intimately familiar with our runners. My kingdom for named arguments in Go...

mkcp (approved):
This looks good! My only review item is nonblocking: we should lean on "string" as the default Format, like hcl, and elide the field in cfg when we're not overriding it.

PR #270, opened 2022-11-11
Add Timeout & Cancelation to Command Runner

nwchandler:
This merge enables timeouts and cancellation in the Command runner. Timeouts may be passed in via optional configuration. If the timeout expires, or if a context that has been passed into the Command runner when it was constructed is canceled, then the run will stop with an appropriate operation status type.

### DIFF
--- a/hcl/hcl.go
+++ b/hcl/hcl.go
@@ -25,8 +25,8 @@ type Blocks interface {
 	*Host | *Product | *Agent
 }
 
-// NOTE(dcohen) this is currently a separate config block, as opposed to a parent block of the others
 type Agent struct {
+	// NOTE(dcohen) this is currently a separate config block, as opposed to a parent block of the others
 	Redactions []Redact `hcl:"redact,block" json:"redactions,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


hcl/hcl.go
```diff
@@ -25,8 +25,8 @@ type Blocks interface {
 	*Host | *Product | *Agent
 }
 
-// NOTE(dcohen) this is currently a separate config block, as opposed to a parent block of the others
 type Agent struct {
+	// NOTE(dcohen) this is currently a separate config block, as opposed to a parent block of the others
```
nwchandler, line 29: The comment directly above an exported object is supposed to be a doc comment, which starts with the name of the object. So, since this one's a note, I pulled it down inside the struct.

groovemonkey (approved):
This is great; appreciated the demo at mid-sprint.

runner/timeout_test.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+)
+
+// ExampleTimeout_MarshalJSON provides some examples to demonstrate the usefulness of the Timeout
+// struct as it relates to marshalling to JSON. Because results from diagnostic runs are written
+// as JSON, this provides a more human-readable format than what the time.Duration type provides.
+func ExampleTimeout_MarshalJSON() {
```
groovemonkey, line 12: baller

mkcp (approved):

PR #271, opened 2022-11-11
Add ConsulDebug Runner

groovemonkey:
Creates an HCL-aware ConsulDebug runner for us to use.


runner/debug/consul.go
```diff
@@ -0,0 +1,101 @@
...
+
+	output string
+}
+
+func (d ConsulDebug) ID() string {
+	return "ConsulDebug"
+}
+
+func NewConsulDebug(cfg ConsulDebugConfig, tmpDir string, debugDuration time.Duration, debugInterval time.Duration) *ConsulDebug {
+	// Create a pseudorandom string of characters to allow >1 ConsulDebug runner without filename collisions
```
nwchandler, line 39: Is this something that could be done with `os.MkdirTemp` instead, similar to what we do with agent.CreateTemp?

runner/debug/consul.go
```diff
@@ -0,0 +1,101 @@
...
+		// Use debug duration and interval
+		Duration: debugDuration.String(),
+		Interval: debugInterval.String(),
+		// Creates a subdirectory inside output dir
+		output:     debugOutputPath(tmpDir, "ConsulDebug", randStr),
+		Captures:   cfg.Captures,
+		Redactions: cfg.Redactions,
+	}
+
+	if len(cfg.Archive) > 0 {
```
nwchandler, line 54: Nit: Since we're not looking for a specific character in a string, but rather, is the string set to something other than `""`, I think this would be more readable as:

```suggestion
	if cfg.Archive != "" {
```

This may just be me, but when I saw the length check, I did a double-take to try and figure out what was happening.

runner/debug/consul.go
nwchandler: When I looked at the `randAlphanumString` function, I noticed that `alphaNumBytes` is missing the number 7; I can't add a comment since that wasn't changed in this PR, but if we use that function still, would you mind taking care of that too?

groovemonkey: Woah, nice! I had no idea os.MkdirTemp can add random patterns -- I'll definitely use that.

runner/debug/consul.go
```diff
@@ -0,0 +1,107 @@
...
+		output:     tmpDir,
+		Captures:   cfg.Captures,
+		Redactions: cfg.Redactions,
+	}
+
+	if cfg.Archive != "" {
+		dbg.Archive = cfg.Archive
+	}
+
+	if len(cfg.Duration) > 0 {
```
nwchandler: Since the other similar checks we've done have compared to the empty string, I think we should do that here too.

nwchandler (approved):

runner/debug/consul.go
nwchandler: I updated this while Dave is away.

PR #273, opened 2022-11-15
Migrate HTTP Runner Constructors to Use Config Objects

nwchandler:
This merge updates the HTTP constructors to leverage a new HttpConfig object, rather than parameters. The intention is to simplify future feature additions by limiting changes to the function signatures of these constructors.

Unrelated to the primary change, a configuration setting required updating for integration tests to pass using a new version of Consul that was recently released, version 1.14.0. This change to Consul does not impact hcdiag, but prevented the dev Consul server from starting during integration tests.

### DIFF
--- a/changelog/273.txt
+++ b/changelog/273.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+runners: Update HTTP runner instantiation to use configuration objects, in order to simplify future feature additions.
+```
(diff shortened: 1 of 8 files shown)


mkcp (approved):
Data-as-system-boundaries & semantic error types, what's not to like? This reminds me that we should schedule some time to rename HTTP to GET. Maybe around `go-fetch` extraction?

PR #274, opened 2022-11-15
Add Timeout & Cancelation to HTTP Runner

nwchandler:
This merge enables timeouts and cancellation in the HTTP runner. Timeouts may be passed in via optional configuration. If the timeout expires, or if a context that has been passed into the runner when it was constructed is canceled, then the run will stop with an appropriate operation status type.

### DIFF
--- a/runner/http.go
+++ b/runner/http.go
@@ -99,12 +99,15 @@ func (h HTTP) Run() op.Op {
 	result := map[string]any{"response": redactedResponse}
 	if err != nil {
 		var failureType op.Status
-		if errors.Is(err, context.DeadlineExceeded) {
+		switch {
+		// The error returned from the http package will wrap timeouts/cancellations, so we check whether we find
+		// these in the error chain to determine the proper op.Status for the failure type.
+		case errors.Is(err, context.DeadlineExceeded):
 			failureType = op.Timeout
-		} else if errors.Is(err, context.Canceled) {
+		case errors.Is(err, context.Canceled):
 			failureType = op.Canceled
-		} else {
-			failureType = op.Fail
+		default:
+			failureType = op.Unknown
 		}
 		return op.New(h.ID(), result, failureType, err, Params(h), startTime, time.Now())
 	}
--- a/runner/http.go
+++ b/runner/http.go
@@ -2,6 +2,7 @@ package runner
 
 import (
 	"context"
+	"errors"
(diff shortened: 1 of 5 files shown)


runner/http.go
```diff
@@ -80,12 +81,32 @@ func (h HTTP) ID() string {
...
+
 	startTime := time.Now()
 
-	redactedResponse, err := h.Client.RedactGet(h.Path, h.Redactions)
+	redactedResponse, err := h.Client.RedactGetWithContext(runCtx, h.Path, h.Redactions)
 	result := map[string]any{"response": redactedResponse}
 	if err != nil {
-		op.New(h.ID(), result, op.Fail, err, Params(h), startTime, time.Now())
+		var failureType op.Status
+		if errors.Is(err, context.DeadlineExceeded) {
```
mkcp, line 102: Can we condense this down into a switch?

```
switch err {
case context.DeadlineExceeded:
...
case context.Canceled:
...
default: 
}
```

runner/http.go
```diff
@@ -80,12 +81,32 @@ func (h HTTP) ID() string {
...
+			failureType = op.Timeout
+		} else if errors.Is(err, context.Canceled) {
+			failureType = op.Canceled
+		} else {
+			failureType = op.Fail
```
mkcp, line 107: I think we ought to classify these catch-all cases as `status.Unknown` and save the `op.Fail` when there's specific errors we can key off of from. Despite the overlap between "fail" and "failed but unknown why" using `Unknown` at least gives us a chance to highlight these cases.

mkcp (approved):
Minor comments, merge when ready 👍

runner/http.go
nwchandler: Good idea! Because the context error types are wrapped by the http package, `errors.Is` is needed, but I updated this code to use a `switch`, which does look cleaner!

nwchandler: I also added a comment to the code to capture the fact that timeouts/cancellations are wrapped by the http package, so that we remember this in the future.

runner/http.go
nwchandler: That makes sense. I think we've gone with `Unknown` in similar cases in other runners. Because the existing code called all errors `op.Fail`, I stuck with that behavior initially. However, I've updated the branch to go with `op.Unknown` now.

PR #275, opened 2022-11-15
WIP: migrate fstab to context and timeouts

mkcp:
The tests are very broken - there's a shorter path to completion once Shell supports timeouts, but there's still an open question about whether DI and mocking out nested-runners is something we want to maintain going forward. The intent is to return to this Runner later in the timeout migration.

### DIFF
--- a/runner/host/fstab.go
+++ b/runner/host/fstab.go
@@ -1,6 +1,7 @@
 package host
 
 import (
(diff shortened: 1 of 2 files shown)


mkcp:
Closing this out because we'll be deferring to the sheller for its timeout.

PR #276, opened 2022-11-15
add timeouts to host/memory

mkcp:
More or less modeled after runner.Command. Tested by setting the default timeout in the host runners to `Timeout(1*time.Nanosecond)` and seeing +1 runner of status.Timeout in the product complete list. We should really fix nested status rendering soon.

### DIFF
--- a/product/host.go
+++ b/product/host.go
@@ -73,6 +73,7 @@ func hostRunners(ctx context.Context, os string, redactions []*redact.Redact, l 
 		host.NewOS(os, redactions),
 		host.NewDisk(redactions),
 		host.NewInfo(redactions),
+		// TODO(mkcp): Source the timeout value from agent/CLI params, or extract to a const.
 		host.NewMemoryWithContext(ctx, runner.Timeout(30*time.Second)),
 		host.NewProcess(redactions),
 		host.NewNetwork(redactions),
--- a/product/host.go
+++ b/product/host.go
@@ -3,6 +3,7 @@ package product
 import (
 	"context"
 	"runtime"
+	"time"
 
 	"github.com/hashicorp/hcdiag/runner/do"
 	"github.com/hashicorp/hcdiag/runner/host"
@@ -72,7 +73,7 @@ func hostRunners(ctx context.Context, os string, redactions []*redact.Redact, l 
 		host.NewOS(os, redactions),
 		host.NewDisk(redactions),
 		host.NewInfo(redactions),
-		host.Memory{},
+		host.NewMemoryWithContext(ctx, runner.Timeout(30*time.Second)),
 		host.NewProcess(redactions),
 		host.NewNetwork(redactions),
 		host.NewEtcHosts(redactions),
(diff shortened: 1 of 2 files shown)


mkcp:
Skipping changelog in this PR in favor of a batch changelog for all host runner timeouts later.

runner/host/memory.go
```diff
@@ -1,32 +1,80 @@
...
+	var runCtx context.Context
+	var cancel context.CancelFunc
+	if 0 < m.Timeout {
+		runCtx, cancel = context.WithTimeout(m.ctx, time.Duration(m.Timeout))
+		defer cancel()
+	}
+
+	go func(resChan chan op.Op, startTime time.Time) {
+		o := m.run()
+		o.Start = startTime
```
nwchandler, line 53: I like this! I wonder, though, if the enclosing `startTime` could be used instead of passing in a parameter to this anonymous function. 🤔

runner/host/memory.go
```diff
@@ -1,32 +1,80 @@
...
+
+	var runCtx context.Context
+	var cancel context.CancelFunc
+	if 0 < m.Timeout {
+		runCtx, cancel = context.WithTimeout(m.ctx, time.Duration(m.Timeout))
+		defer cancel()
+	}
+
+	go func(resChan chan op.Op, startTime time.Time) {
+		o := m.run()
```
nwchandler, line 52: Nice! I'm thinking about going back and doing this with Command now. 😄

product/host.go
```diff
@@ -72,7 +73,7 @@ func hostRunners(ctx context.Context, os string, redactions []*redact.Redact, l
 		host.NewOS(os, redactions),
 		host.NewDisk(redactions),
 		host.NewInfo(redactions),
-		host.Memory{},
+		host.NewMemoryWithContext(ctx, runner.Timeout(30*time.Second)),
```
nwchandler, line 76: Not a big deal here, but if we end up making 30 seconds a common default across host runners, I think it may be worth making it a constant so that it's easier to update it.

nwchandler (approved):
Looks good to me! I left a few minor comments/thoughts, but nothing to hold up approval.

runner/host/memory.go
```diff
@@ -1,32 +1,80 @@
...
 	"github.com/hashicorp/hcdiag/runner"
-	"github.com/shirou/gopsutil/v3/mem"
 )
 
 var _ runner.Runner = Memory{}
 
-type Memory struct{}
+type Memory struct {
+	ctx     context.Context
+	Timeout runner.Timeout
```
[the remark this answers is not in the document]
mkcp: Yes, fixed in revision

runner/host/memory.go
mkcp: Shadowing the var name to pass it into the closure is a bad practice, but copying the startTime to avoid modification helps a little bit with system robustness. I shared some implementation ideas out of band for an immutable `Start` type for `Op.Start`, but I'm going to kick that can down the road until a genuine need arises.

product/host.go
mkcp: I'll make a todo note as we keep implementing timeouts in the runners. Ideally the defaults would be sourced from the product via the agent, but for any one-offs then a const is for sure the way to go.

PR #277, opened 2022-11-16
Add timeouts to shell runner

mkcp:
Adds context and timeout support to `runner.Shell`. Includes HCL support and updates to all dependencies calling `runner.Shell`.

### DIFF
--- a/runner/shell.go
+++ b/runner/shell.go
@@ -1,6 +1,7 @@
 package runner
 
 import (
+	"context"
 	"fmt"
 	"os/exec"
 	"time"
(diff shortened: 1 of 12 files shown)


runner/shell_test.go
```diff
@@ -18,7 +18,10 @@ func TestShell(t *testing.T) {
 	os.Setenv("SHELL", "/bin/sh")
 
 	// features pipe "|" and file redirection ">"
-	c := NewShell("echo hiii | grep hi > cooltestfile", nil)
```
nwchandler, line 21: Should we include some timeout/cancellation tests?

runner/shell.go
```diff
@@ -12,19 +13,44 @@ import (
...
+
+// NewShellWithContext provides a runner for arbitrary shell code.
+func NewShellWithContext(ctx context.Context, cfg ShellConfig) (*Shell, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
+	timeout := cfg.Timeout
+	if timeout < 0 {
+		return nil, fmt.Errorf("timeout must be a nonnegative, timeout='%s'", timeout.String())
```
nwchandler, line 45: I wouldn't block for this, but in Command and HTTP, I included config error types, which I thought could be useful for upstream type checking. Should we do the same for `Shell`? I would prefer if the foundational runners were consistent in this way, either with custom types or with a `fmt.Errorf`.

nwchandler (approved):
Left a couple comments for consideration, but this generally looks good to me!

runner/shell.go
mkcp: I started implementing a `ShellConfigError` in this PR, but with just the timeout validation decided to skip the type. We definitely should go with a custom type in the future when there's more config validation here and/or we're specifically checking for Config error across multiple types of runners.

PR #278, opened 2022-11-21
Add EtcHosts + Bugfixes and helpers

mkcp:
This PR adds config, ctx, and timeouts to `EtcHosts` by passing through to the Shell runner it wraps. As I was checking that the timeouts worked correctly, I found that we were getting back an empty error from the wrapped Shell. This is because we were returning the error on the wrong context in the op return. This bug also existed on Command so I addressed it there.

Alongside this, I was experimenting with some helper functions that fill in the commonly used fields in Timeout and Cancel status ops, so those are also deployed here.

### DIFF
--- a/op/op.go
+++ b/op/op.go
@@ -61,10 +61,12 @@ func New(id string, result map[string]any, status Status, err error, params map[
 	}
 }
 
+// NewCancel takes the minimum required fields to build and return an Op with status Canceled
 func NewCancel(id string, err error, params map[string]any, start time.Time) Op {
 	return New(id, map[string]any{}, Canceled, err, params, start, time.Now())
 }
 
+// NewTimeout takes the minimum required fields to build and return an Op with status Timeout
 func NewTimeout(id string, err error, params map[string]any, start time.Time) Op {
 	return New(id, map[string]any{}, Timeout, err, params, start, time.Now())
 }
--- a/op/op.go
+++ b/op/op.go
@@ -61,6 +61,14 @@ func New(id string, result map[string]any, status Status, err error, params map[
 	}
 }
 
+func NewCancel(id string, err error, params map[string]any, start time.Time) Op {
+	return New(id, map[string]any{}, Canceled, err, params, start, time.Now())
+}
+
+func NewTimeout(id string, err error, params map[string]any, start time.Time) Op {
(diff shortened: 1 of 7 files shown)


op/op.go
```diff
@@ -61,6 +61,14 @@ func New(id string, result map[string]any, status Status, err error, params map[
 	}
 }
 
+func NewCancel(id string, err error, params map[string]any, start time.Time) Op {
```
nwchandler, line 64: Minor: To help prep for library extract, it may be worth adding a doc comment to this and `NewTimeout` so we don't have to go back and add them later.

product/host.go
```diff
@@ -16,6 +16,11 @@ import (
 	"github.com/hashicorp/hcdiag/runner"
 )
 
+const (
+	TimeoutTenSeconds    = runner.Timeout(10 * time.Second)
+	TimeoutThirtySeconds = runner.Timeout(30 * time.Second)
```
nwchandler, line 21: I wonder if we should have a single default timeout for host runners or if we actually need multiple. And, perhaps we could give it a semantic name, so it was clearer why it is, rather than what it is. Perhaps something like:
```suggestion
	DefaultHostTimeout    = runner.Timeout(30 * time.Second)
```

runner/command.go
```diff
@@ -193,11 +193,11 @@ func (c Command) Run() op.Op {
 	case <-runCtx.Done():
 		switch runCtx.Err() {
 		case context.Canceled:
-			return op.New(c.ID(), nil, op.Canceled, c.ctx.Err(), Params(c), startTime, time.Now())
+			return op.New(c.ID(), nil, op.Canceled, runCtx.Err(), Params(c), startTime, time.Now())
```
nwchandler, line 196: Good catch! 👍

nwchandler (approved):
I left a couple minor comments for consideration, but otherwise, LGTM!

product/host.go
mkcp: I want to come back to potential shared default timeouts once we've added a bunch of runner-specific defaults and know more about each runner. Using a longer shared time is more what i would expect from a deadline set at the agent or product context. Having tighter runner-specific timeouts makes retries with timeouts possible.

PR #279, opened 2022-11-22
Update runner.Params() to use any

mkcp:
Noticed this cleanup while working on the timeout and cancel op helpers.


nwchandler (approved):
👍

PR #281, opened 2022-11-30
Add Timeout & Cancelation to Docker Log Runner

nwchandler:
This merge enables timeouts and cancellation in the Docker Log runner. Timeouts may be passed in via optional configuration. If the timeout expires, or if a context that has been passed into the runner when it was constructed is canceled, then the run will stop with an appropriate operation status type.


mkcp (approved):
LGTM!

PR #282, opened 2022-12-02
Traverse Nested Runners for ManifestOps

nwchandler:
This merge addresses an issue with reporting that was introduced when product runners were wrapped inside Do blocks. Previous reporting, both in manifest.json and in the end-of-run results summary, did not account for nested runners, so they would only show 1 runner (the Do runner) instead of indicating all of the nested runner counts. This is now addressed, with a new function that walks arbitrarily nested Op maps and produces a slice of ManifestOp that incorporates each one.


mkcp (approved):
Thanks for pairing and carrying this across the finish line!

PR #289, opened 2022-12-09
Add Timeout & Cancellation to Host/Info Runner

nwchandler:
This merge enables timeouts and cancellation in the host info runner. Timeouts may be passed in via optional configuration. If the timeout expires, or if a context that has been passed into the runner when it was constructed is canceled, then the run will stop with an appropriate operation status type.

### DIFF
--- a/product/host.go
+++ b/product/host.go
@@ -80,7 +80,7 @@ func hostRunners(ctx context.Context, os string, redactions []*redact.Redact, l 
 	r := []runner.Runner{
 		host.NewOS(os, redactions),
 		host.NewDiskWithContext(ctx, host.DiskConfig{Redactions: redactions}),
-		host.NewInfo(redactions),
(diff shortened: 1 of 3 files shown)


product/host.go
```diff
@@ -80,7 +80,7 @@ func hostRunners(ctx context.Context, os string, redactions []*redact.Redact, l
 	r := []runner.Runner{
 		host.NewOS(os, redactions),
 		host.NewDiskWithContext(ctx, host.DiskConfig{Redactions: redactions}),
-		host.NewInfo(redactions),
+		host.NewInfoWithContext(ctx, host.InfoConfig{Redactions: redactions}),
```
mkcp, line 83: Thoughts on a default of `timeout: TimeoutTenSeconds` ?

mkcp (approved):
LGTM, non-blocking comment about passing in a default timeout

PR #293, opened 2022-12-14
Deprecate Debug Duration and Debug Interval Flags & Document Their Alternatives

nwchandler:
This merge marks the -debug-duration and -debug-interval flags as deprecated, as we want to encourage users to leverage custom configuration. To that end, the merge also provides documentation about the new product-specific debug runners' configuration via HCL.


mkcp (approved):
Excellent, thank you for taking this on!

PR #295, opened 2023-01-03
Add Timeout & Cancellation to Host OS Runner

nwchandler:
This merge enables timeouts and cancellation in the Host OS runner, with a default of 10 seconds. If the timeout expires, or if a context that has been passed into the runner when it was constructed is canceled, then the run will stop with an appropriate operation status type.


mkcp (approved):
👍

PR #297, opened 2023-01-09
Fix Empty Timeout Error in HCL Shell Runner

nwchandler:
This merge addresses an issue where the timeout parameter was essentially required for the Shell runner in HCL. Without it being set, an error was returned in converting nil to a duration.


mkcp (approved):
Thanks for coaxing this bug onto a piece of paper and gently releasing it outside

PR #298, opened 2023-01-17
Add fstab timeouts and context

mkcp:
This PR adds an FSTab config struct with a Timeout field and adds a context constructor. The timeout value passes into the Shell runner and propages the `op.Status` and `op.Error` from the result op.


nwchandler (approved):
LGTM!

PR #302, opened 2023-01-25
Add Additional Vault Runners

nwchandler:
Vault now includes additional runners for troubleshooting, from the following API endpoints: sys/audit, sys/version-history?list=true, sys/license/status, sys/replication-status.


mkcp (approved):
Awesome, thank you for taking this on. LGTM!

PR #308, opened 2023-01-30
Disable CGO When Building Binaries in GitHub

nwchandler:


mkcp (approved):
Change looks good, as discussed in slack 👍

Issue #310, opened 2023-01-30
bash: /usr/share/keyrings/hashicorp-archive-keyring.gpg: Permission denied


MikaelHashiCorp:
**What is the bug?**  
When I try to install hcdiag from the [installation instructions](https://github.com/hashicorp/hcdiag/blob/main/docs/installation.md) for [Example: Ubuntu](https://github.com/hashicorp/hcdiag/blob/main/docs/installation.md), running the command:

```
wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor > /usr/share/keyrings/hashicorp-archive-keyring.gpg && echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com jammy main" | tee /etc/apt/sources.list.d/hashicorp.list
```

I get the following error:

```
bash: /usr/share/keyrings/hashicorp-archive-keyring.gpg: Permission denied
```

Full message:

```
ubuntu@ip-172-31-22-248:~/hcd-uat$ wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor > /usr/share/keyrings/hashicorp-archive-keyring.gpg && echo "deb [signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com jammy main" | tee /etc/apt/sources.list.d/hashicorp.list

bash: /usr/share/keyrings/hashicorp-archive-keyring.gpg: Permission denied
--2023-01-30 22:30:46--  https://apt.releases.hashicorp.com/gpg
Resolving apt.releases.hashicorp.com (apt.releases.hashicorp.com)... 18.160.46.62, 18.160.46.37, 18.160.46.7, ...
Connecting to apt.releases.hashicorp.com (apt.releases.hashicorp.com)|18.160.46.62|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 3980 (3.9K) [binary/octet-stream]
Saving to: ‘STDOUT’

-                                          0%[                                                                                 ]       0  --.-KB/s    in 0s      


Cannot write to ‘-’ (Success).
```

**What version are you using?**  
Trying to install hcdiag version [hcdiag_0.5.0-rc1](https://releases.hashicorp.com/hcdiag/0.5.0-rc1/)

**Which HashiCorp products does this affect?**  
hcdiag

**When does the error occur?**  
1. Spin up an instance of Ubuntu 20.04.5 LTS.
[...]

PR #312, opened 2023-01-31
add note about ubuntu installation requiring root privileges

groovemonkey:
It also demonstrates switching to the root user in the example script.

RE: https://github.com/hashicorp/hcdiag/issues/310


changelog/288.txt
```diff
@@ -1,3 +1,3 @@
 ```release-note:improvement
-runners: The host IPTables runner now takes an optional Timeout value, so that long-running executions can be gracefully stopped.
+documentation: add note about ubuntu installation requiring root privileges, and demonstrate switching user in the example commands
```
nwchandler, line 2: I think maybe there was a file copied but not renamed. 😄

nwchandler (approved):
Approved (once the changelog is updated)

changelog/288.txt
groovemonkey: Sorry, I *severely* mis-read a directory listing. Twice.

PR #325, opened 2023-05-18
Include hcdiag-ext in core of hcdiag

roooms:
Requesting we add the checks from hcdiag-ext into hcdiag "core" since these are very lightweight and would reduce customer confusion and complexity when we ask customers to run the hcdiag-ext workflow today. This would allow customers to just run hcdiag if they find that easier.

### DIFF
--- a/changelog/325.txt
+++ b/changelog/325.txt
@@ -1,0 +1,3 @@
+```release-note:improvement
+seeker: add hcdiag-ext seekers/runners into hcdiag core due to lightweight improvement and benefit to customer experience
+```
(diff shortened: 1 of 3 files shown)


roooms:
Please can this be re-opened and @menicosia provide updates here? This is an important inclusion we would like to understand the progress and decision process.

menicosia:
Acknowledge. I will check with the team if it's possible to re-open a PR that's already been marked as Merged. If and only if that isn't possible, I will work offline with the deltas provided here, get feedback, and provide that in asking you for a new PR.

PR #327, opened 2023-05-26
Fix replicated support bundle generation command

zisom-hc:
The command as it stands currently does not work without the inclusion of an App ID. Instead of having to obtain and contain the App ID into a variable, that is then passed to this command, we can instead execute the command we in support typically execute, which is `replicatedctl support-bundle`

Current results of that command when executing `hcdiag -terraform-ent`:

```
"replicated support-bundle": {
                            "result": {
                                "text": "Error: Usage: replicated support-bundle \u003capp id\u003e\n"
                            },
```

What the output is if you tried to execute the command manually within a TFE installation:

```
root@ip-10-0-166-24:/var/snap/amazon-ssm-agent/6312# replicated support-bundle
Error: Usage: replicated support-bundle <app id>
```

### DIFF
--- a/changelog/327.txt
+++ b/changelog/327.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+TFE: Change replicated support bundle generation command to avoid the necessity for an app ID variable to be provided.
+```
+
--- a/product/tfe.go
+++ b/product/tfe.go
@@ -72,7 +72,7 @@ func tfeRunners(ctx context.Context, cfg Config, api *client.APIClient, l hclog.
 
 	// Set up the Replicated Support Bundle runners
 	supportBundleCmd, err := runner.NewCommandWithContext(ctx, runner.CommandConfig{
-		Command:    "replicated support-bundle",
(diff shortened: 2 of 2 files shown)


aliciaclark1066 (approved):

aliciaclark1066:
@zisom-hc Thanks for reporting! We'd add a change log and get another release cut.

@nwchandler Can you add the change log comment here?

nwchandler:
@zisom-hc We'll get this into a hcdiag 0.5.1 release. Is it okay if that release happens next week, probably on Wednesday (due to Monday holiday and PTO for me on Tuesday)?

zisom-hc:
> @zisom-hc We'll get this into a hcdiag 0.5.1 release. Is it okay if that release happens next week, probably on Wednesday (due to Monday holiday and PTO for me on Tuesday)?

Hey Nick, yeah that's all good with me.

nwchandler:
Sounds good! I'll get the release process kicked off Wednesday morning and let you know once it's ready.

PR #329, opened 2023-07-07
Remove Serial Execution Option

nwchandler:
This merge removes the `-serial` flag and the supporting code around it. When concurrent runners were added, serial only caused semi-serial execution, so the flag was deprecated in 0.5.0. This merge actually removes it in preparation for 0.6.0.


groovemonkey (approved):
Haven't built/tested this but it looks good to me!

PR #331, opened 2023-07-10
Remove Includes Flag

nwchandler:
The -includes flag was deprecated in v0.5.0. This merge removes it. Users who need to copy files should, instead, use configuration files with Copy runners.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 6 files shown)


agent/agent.go
```diff
@@ -20,7 +19,6 @@ import (
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/hcdiag/product"
-	"github.com/hashicorp/hcdiag/runner"
```
groovemonkey, line 23: Oh, that's a neat side effect. Didn't know that includes were the only thing left coupling the Agent to Runners.

nwchandler: Yeah, I was surprised too!

Ayeyiiat: I learnt something new, nice!

PR #332, opened 2023-07-10
add a docker development environment

groovemonkey:
Virtualbox is no longer an option on Apple Silicon, and that makes our vagrant environment harder to set up for devs new to working on hcdiag. I've created a basic docker development environment that makes it easy to test hcdiag in a docker container.

This container setup uses the same base image and install process that our learn tutorials recommend (I came up with that too, so if you don't like how this is done I'm happy to hear alternatives :-D), so it'll help keep code rot from setting in there as well.

## Developer Experience:
Developers with docker-engine installed can run the following commands to start testing hcdiag in a container:

1. Build and start the container: `make compose-up`
1. Compile hcdiag for linux (automatically shows up in the container via the directory mount): `GOOS=linux GOARCH=amd64 go build -o bin/hcdiag`
1. Connect to the container in a new shell: `docker exec -it hcdiag-hashistack-1 /bin/bash`
1. Move to the working directory and source the env file: `cd /hashistack && source hashi.env`
1. Run your new hcdiag build: `/hcdiag/hcdiag`

The container's init script reminds you of these steps each time so there's nothing to remember:

Screenshot 2023-07-10 at 3 22 13 PM

## Implementation

1. There's a setup script (`docker-init.sh`) which configures Ubuntu, installs Nomad, Vault, and Consul, creates an .env file so the product binaries "just work," starts the products in dev mode, and sleeps.
1. There's a `Dockerfile` which copies a setup script to the base image. I've kept this simple so that the script can do all the heavy lifting at runtime, too (running services and then blocking forever so devs can `docker exec` in and work).
1. There's a docker-compose file which builds the container and mounts the hcdiag `bin` directory in the container at /hcdiag.
1. There's an additional target in our Makefile, `docker-compose`.

## Feedback
I'm not married to this idea or to this exact approach, I just started working on hcdiag this morning and wanted something that works. If you have suggestions for improvements, please leave a comment here!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM ubuntu:22.04
+COPY docker-init.sh /usr/local/bin/
+
+# Entrypoint will be set by docker-compose:"command"
+ENTRYPOINT []
+
--- a/Makefile
+++ b/Makefile
@@ -55,5 +55,9 @@ version: ## Show the version of the project
 	@echo $(VERSION)
 .PHONY: version
 
+compose-up: # run docker compose in the foreground
+	docker-compose up --force-recreate --build --remove-orphans
+.PHONY: compose-up
+
 # windows:
 # $env:path = "$pwd/bin;$env:path"
--- a/changelog/332.txt
+++ b/changelog/332.txt
@@ -1,0 +1,4 @@
+```release-note:improvement
+tooling: create a docker-based development environment.
+```
+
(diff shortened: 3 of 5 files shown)


Ayeyiiat:
Based on how it compares with vagrant+qemu, this is magic!

Ayeyiiat (approved):
Looks good to me.

PR #334, opened 2023-07-11
multiplexing output to a logfile and stdout

groovemonkey:
This includes the necessary refactoring needed to enable this feature.

### Refactoring
- Moved .tar.gzipping out of the Agent and into the `utils` package. It's now called from the cli (`run.go`).
- Added a MultiWriter to let us multiplex output.
- Rewrote tempfile creation to allow this, and moved it out of the Agent and into the `utils` package.
- Tempdir path and logfile for the hcdiag run are now created in the cli (`run.go`). Tempdir location is passed into the Agent with a new Agent.Config field.
- Deprecated Agent.TmpDir() and Agent.resultsDest -- they were more or less unused

### TODO:
- log coloring is currently broken (via the config in run.go). I'll try to come up with a clever way around this -- maybe I'll write a logging wrapper that takes the `MultiWriter` and writes to all writers it contains, in addition to making regular hclog calls to stdout (with coloring).
- Tests for moved functions (createTemp stuff should be moved from `agent_test.go` to `util/util_test.go`).
- A thorough functional look to make sure this all works. I've tested it in a docker container and it behaves how I'd expect so far.
- Fix integration tests.

This PR implements the work outlined in https://hashicorp.atlassian.net/browse/CORI2-188


command/run.go
```diff
@@ -144,6 +145,44 @@ func (c *RunCommand) Synopsis() string {
...
+	var err error
+
+	// Default teeWriter just prints to stdout
+	var teeWriter io.Writer = os.Stdout
+
+	// Create a temporary directory and logfile
+	if !c.dryrun {
+		if tmp, err = util.CreateTemp(); err != nil {
+			fmt.Println("Failed to create temp directory. error:", err)
+			return 1
```
nwchandler, line 158: Should we return a return code here instead so we can name it?

agent/agent.go
```diff
@@ -36,6 +36,7 @@ type Config struct {
 	Since       time.Time `json:"since"`
 	Until       time.Time `json:"until"`
 	Destination string    `json:"destination"`
+	TmpDir      string    `json:"tmpdir"`
```
nwchandler, line 39: Nit:
```suggestion
	TmpDir      string    `json:"tmp_dir"`
```

agent/agent.go
```diff
@@ -100,6 +100,11 @@ func NewAgentWithContext(ctx context.Context, config Config, logger hclog.Logger
 		redacts = redact.Flatten(hclRedacts, redacts)
 	}
 
+	// Ensure temporary directory exists
+	if _, err = os.Stat(config.TmpDir); err != nil {
+		return nil, fmt.Errorf("Error creating tmpdir for new agent: %w", err)
```
nwchandler, line 105: Should the error say that the temporary directory doesn't exist, rather than saying it failed to create? It looks like we aren't creating it here, so the error might be misleading.

agent/agent.go
```diff
@@ -337,15 +294,9 @@ func (a *Agent) RecordManifest() {
 	a.ManifestOps = result
 }
 
-// WriteOutput renders the manifest and results of the diagnostics run and writes the compressed archive.
+// WriteOutput renders the manifest and results of the diagnostics run into Agent.tmpdir
```
nwchandler, line 297: Nit:
```suggestion
// WriteOutput renders the manifest and results of the diagnostics run into Agent.TmpDir
```

agent/agent.go
```diff
@@ -337,15 +294,9 @@ func (a *Agent) RecordManifest() {
...
 func (a *Agent) WriteOutput() (err error) {
-	err = util.EnsureDirectory(a.Config.Destination)
-	if err != nil {
-		a.l.Error("Failed to ensure destination directory exists", "dir", a.Config.Destination, "error", err)
-		return err
-	}
-
-	a.l.Debug("Writing results and manifest, and creating tar.gz archive")
+	a.l.Debug("Writing results and manifest")
```
nwchandler, line 299: Since this is a debug, maybe we should include the location in it - that may help users who are troubleshooting some issue or other, if they don't know where the temp directory is already.

util/util.go
```diff
@@ -210,6 +210,20 @@ func FindInInterface(iface interface{}, mapKeys ...string) (interface{}, error)
 	return mapped, nil
 }
 
+// CreateTemp Creates a temporary directory so that we may gather results and files before compressing the final
+// artifact. It returns a string representing a filepath, and an optional error.
+func CreateTemp() (string, error) {
+	tmp, err := os.MkdirTemp(os.TempDir(), "hcdiag")
```
nwchandler, line 216: I think this is a fairly big change from current behavior, as we are putting the temp directory in the OS's temp directory instead of the destination directory, and we don't include the timestamp piece of the directory name. I wonder if we should keep similar behavior in this particular change. 🤔

command/run.go
```diff
@@ -144,6 +145,44 @@ func (c *RunCommand) Synopsis() string {
...
+		}
+
+		// remove the temporary directory and its contents
+		defer func() {
```
nwchandler, line 162: OK, so this might be a little weird, but what if CreateTemp returned a cleanup function as well. So, we had something like

```go
if tmp, cleanup, err = util.CreateTemp(); err != nil {
// ...
}

defer cleanup()
```

command/run.go
```diff
@@ -208,22 +251,49 @@ func (c *RunCommand) Run(args []string) int {
...
 // configureLogging takes a logger name, sets the default configuration, grabs the LOG_LEVEL from our ENV vars, and
 // returns a configured and usable logger.
-func configureLogging(loggerName string) hclog.Logger {
+func configureLogging(loggerName string, loggerOutput io.Writer) hclog.Logger {
 	// Create logger, set default and log level
 	appLogger := hclog.New(&hclog.LoggerOptions{
-		Name:  loggerName,
-		Color: hclog.AutoColor,
+		Name: loggerName,
+		// Color:  hclog.AutoColor,
```
nwchandler, line 295: Looks like some commented code

nwchandler (commented):
Looks like a nice approach! I left some comments inline where I had thoughts.

command/run.go
groovemonkey: Love that; very clever.

command/run.go
groovemonkey: Yeah, this change currently breaks coloring but today I'll look at wrapping this logger in our logging functions to fix that.

agent/agent.go
groovemonkey: Just to be sure: your intent is to export the temporary directory field on the agent?

nwchandler: Oh, I was just thinking that the name of the field wasn't "tmpdir," but it's "TmpDir." Apologies if I misunderstood!

command/run.go
nwchandler: Cool! I learned it from tests that set up temporary stuff that needs cleaned up, but it seemed to fit here too!

command/run.go
nwchandler: Ah! That makes sense.

util/util.go
nwchandler: Good point about the agent start time. I hadn't considered that!

agent/agent.go
groovemonkey: This is a bit weird -- `agent.Config.TmpDir` is new (and exported, so people can work with it to set up a new Agent) but `Agent.tmpDir` is not -- I left it that way because I wasn't sure we wanted people messing with an existing agent's temporary directory (they should do that through the `agent.Config` type).

If that's too confusing (or wrong) I'm happy to change it!

nwchandler: Ah, I think I looked at the wrong struct when I glanced back up. I think the comment should be `tmpDir` not `tmpdir` (since the "d" is capitalized in the field name and Go is case sensitive).

util/util.go
```diff
@@ -210,6 +210,33 @@ func FindInInterface(iface interface{}, mapKeys ...string) (interface{}, error)
...
+		return "", emptyCleanup, fmt.Errorf("Error creating temp directory, message=%w", err)
+	}
+	tmp, err = filepath.Abs(tmp)
+	if err != nil {
+		return "", emptyCleanup, fmt.Errorf("Error identifying absolute path for temp directory, message=%w", err)
+	}
+
+	// return our temporary directory path, cleanup function, and error
+	return tmp,
+		func(l hclog.Logger) {
```
nwchandler, line 231: Neat way of getting a logger in here!

command/run.go
```diff
@@ -144,6 +145,41 @@ func (c *RunCommand) Synopsis() string {
...
+	if !c.dryrun {
+		if tmp, cleanup, err = util.CreateTemp(c.destination); err != nil {
+			fmt.Println("Failed to create temp directory. error:", err)
+			return SetupError
+		}
+		// remove the temporary directory and its contents
+		defer cleanup(l)
+
+		// Set up stdout/logfile output
+		logfile, err := os.Create(path.Join(tmp, "hcdiag.log")) // creates a file at current directory
```
nwchandler, line 167: Would `filepath.Join` be safer here for Windows machines?

nwchandler (approved):
Left one comment about filepath vs. path, but this looks good to me!

command/run.go
groovemonkey: D'oh! Yes, thank you. Made myself a nice windows bug in another place where I mixed up path/filepath.

util/util.go
groovemonkey: I debated so many different things for this -- first I had the cleanup function return an error, but then we couldn't really `defer cleanup()` anymore without triggering golangci-lint (unhandled error return). Handling it everywhere made things extremely messy, but I'd still like a chance to report on a cleanup failure, since that could affect a customer. Passing in a logger was my compromise. 🤷

PR #335, opened 2023-07-12
Remove Debug-Interval and Debug-Duration Flags

nwchandler:
The -debug-duration and debug-interval flags were deprecated in v0.5.0. This merge removes them. Users who need to adjust debug timers should, instead, use custom configuration files with product debug runners.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -415,11 +415,13 @@ func (a *Agent) Setup() error {
 
 	// Create the base config that we copy into each product
 	baseCfg := product.Config{
-		TmpDir:     a.tmpDir,
-		Since:      a.Config.Since,
-		Until:      a.Config.Until,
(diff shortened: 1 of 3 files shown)


tehut (approved):
This looks like a relatively straight forward deletion PR.  

Do you think it might be useful to add something about debuggers to https://github.com/hashicorp/hcdiag/blob/main/docs/custom-config.md ?

There is also a reference to this flag on[ ln 144](https://github.com/hashicorp/hcdiag/blob/main/docs/custom-config.md?plain=1#L144) that you may want to remove with this PR but I don't think it's blocking.

groovemonkey (approved):

nwchandler:
After some internal discussion, I'm going to close this PR for the time being. There are some potential edge cases around removing these flags that will require further evaluation.

PR #336, opened 2023-07-13
Added the lsmod runner

Ayeyiiat:


runner/host/lsmod_test.go
```diff
@@ -0,0 +1,139 @@
+// Copyright (c) HashiCorp, Inc.
```
nwchandler, line 1: Please be sure to include the SPDX line commented above in lsmod.go here as well.

runner/host/lsmod_test.go
```diff
@@ -0,0 +1,139 @@
...
+			assert.Equal(t, expected.result, o.Result)
+			assert.Equal(t, expected.status, o.Status)
+			if tc.expected.expectErr {
+				assert.Error(t, o.Error)
+			}
+		})
+	}
+}
+
+// Giving a sample test to be run.
```
nwchandler, line 113: I'm not sure if this comment adds too much to readability. Since it said sample, my first thought was that this was just a reference, but it appears to be a "normal" test. I'd maybe prefer just removing the comment.

runner/host/lsmod_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/hcdiag/op"
+
+	"github.com/hashicorp/hcdiag/runner"
+	"github.com/stretchr/testify/assert"
+)
+
+type sampleShellRunner struct {
```
nwchandler, line 14: I wouldn't hold up merging this PR, but I had an idea. It looks like this is identical to the `mockShellRunner` in fstab_test.go. Technically, because these files are in the same package, the lsmod_test.go file could use that same mockShellRunner without having to duplicate it and give it a new name in this file. However, since it is shared, I wonder if we should move the mockShellRunner out of fstab_test.go and into host_test.go, and then it would just be a shared struct type that any test in the package could use.

nwchandler (commented):
This looks good to me! I had a few comments that I think should be addressed, but it's a good PR!

runner/host/lsmod_test.go
Ayeyiiat: Yes that's exactly what it is. I can try to implement that and make sure it works.

Ayeyiiat: I made a new commit with the recommended changes if you wanna take a look.

runner/host/fstab_test.go
```diff
@@ -9,29 +9,10 @@ import (
...
-		Error:  m.err,
-	}
-}
-
-func (m mockShellRunner) ID() string {
-	return ""
-}
-
-var _ runner.Runner = mockShellRunner{}
+//var _ runner.Runner = mockShellRunner{}
```
nwchandler, line 15: Could we move this declaration over to host_test.go as well?

nwchandler (approved):
Looks good to me! I noticed one line was commented out in fstab that I think should probably be moved over to host_test.go as well, but otherwise 👍 !

runner/host/fstab_test.go
Ayeyiiat: Oh yeah I actually meant to remove it. What exactly is it doing because without it, the program still runs

groovemonkey: I had the same question when I started working on this project! This is a funky little trick to make sure we get a compile error if `mockShellRunner` ever stops implementing the runner.Runner interface. Basically, if we accidentally break something that's required to satisfy the `Runner` interface that the rest of hcdiag depends on, this will give a clear error when you try to `go build`.

groovemonkey (approved):
Woohoo, your first PR! Nice job.

runner/host/fstab_test.go
Ayeyiiat: Nice!

PR #337, opened 2023-07-17
Added the Dmesg runner

Ayeyiiat:
Captures the output of the dmesg-T command on linux OS'es.

### DIFF
--- a/product/host.go
+++ b/product/host.go
@@ -124,6 +124,17 @@ func hostRunners(ctx context.Context, os string, redactions []*redact.Redact, l 
 	}
 	r = append(r, lsMod)
 
+	//Dmesg runner
+	dMesg, err := host.NewDMesg(host.DMesgConfig{
+		OS:         os,
+		Timeout:    TimeoutTenSeconds,
+		Redactions: redactions,
+	})
+	if err != nil {
+		l.Error("unable to create host.Dmesg runner.", "err=", err)
+	}
+	r = append(r, dMesg)
+
 	runners := []runner.Runner{
 		do.New(l, "host", "host runners", r),
 	}
--- a/runner/host/dmesg.go
+++ b/runner/host/dmesg.go
@@ -6,6 +6,7 @@ package host
 import (
 	"context"
 	"fmt"
+	"runtime"
 	"time"
 
 	"github.com/hashicorp/hcdiag/op"
@@ -41,6 +42,21 @@ func NewDMesgWithContext(ctx context.Context, cfg DMesgConfig) (*DMesg, error) {
 	if ctx == nil {
 		ctx = context.Background()
 	}
+	if runtime.GOOS == "darwin" {
+		shell, err := runner.NewShellWithContext(ctx, runner.ShellConfig{
+			Command:    "sudo dmesg",
+			Redactions: cfg.Redactions,
+			Timeout:    time.Duration(cfg.Timeout),
+		})
+		if err != nil {
+			return nil, err
+		}
+		return &DMesg{
+			OS:    cfg.OS,
+			Shell: shell,
+			ctx:   ctx,
+		}, nil
+	}
 	shell, err := runner.NewShellWithContext(ctx, runner.ShellConfig{
 		Command:    "dmesg -T",
 		Redactions: cfg.Redactions,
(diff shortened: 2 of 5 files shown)


nwchandler (approved):
As discussed offline, it may be worth adding support for Darwin/Mac too, but this looks good to me.

runner/host/dmesg.go
```diff
@@ -41,6 +42,21 @@ func NewDMesgWithContext(ctx context.Context, cfg DMesgConfig) (*DMesg, error) {
 	if ctx == nil {
 		ctx = context.Background()
 	}
+	if runtime.GOOS == "darwin" {
+		shell, err := runner.NewShellWithContext(ctx, runner.ShellConfig{
+			Command:    "sudo dmesg",
```
nwchandler, line 47: We generally require users to run hcdiag with permissions get run any of the commands it requires, so I would leave `sudo` out of this command and just make it `dmesg`.

runner/host/dmesg.go
```diff
@@ -41,6 +42,21 @@ func NewDMesgWithContext(ctx context.Context, cfg DMesgConfig) (*DMesg, error) {
 	if ctx == nil {
 		ctx = context.Background()
 	}
+	if runtime.GOOS == "darwin" {
```
nwchandler, line 45: The `cfg` has an OS field in it, and we should use it here instead `runtime.GOOS` I think. We pass the OS into the host runners to make some testing scenarios a bit easier.

nwchandler, line 45: I think the flow of logic here could be simplified a bit. In essence, the only difference between darwin and Linux here is the `Command` field on the `runner.ShellConfig`. I would propose we create a `runner.ShellConfig` with the Linux settings. Then, we do a quick `if cfg.OS == "darwin"` check, and if it's true, we set the command to `dmesg` instead of `dmesg -T`.

Then, we can pick back up with the `shell, err := runner.NewShellWithContext(ctx, <the name of the shell config variable created in the previous paragraph> ...`

runner/host/dmesg.go
```diff
@@ -58,15 +74,18 @@ func NewDMesgWithContext(ctx context.Context, cfg DMesgConfig) (*DMesg, error) {
 
 // ID returns the runner ID for DMesg.
 func (r DMesg) ID() string {
+	if runtime.GOOS == "darwin" {
```
nwchandler, line 77: Similar to the above, here we should use `r.OS` instead of `runtime`.

runner/host/dmesg.go
```diff
@@ -0,0 +1,93 @@
...
+var _ runner.Runner = DMesg{}
+
+// DMesgConfig takes each parameter to configure an DMesg runner implementation.
+type DMesgConfig struct {
+	OS         string           `json:"os"`
+	Timeout    runner.Timeout   `json:"timeout"`
+	Redactions []*redact.Redact `json:"redactions"`
+}
+
+// DMesg accepts
```
nwchandler, line 27: I forgot to comment earlier, but it looks like this comment is missing some detail.

Ayeyiiat:
I think I capture a better flow of the logic in the new commit.

product/host.go
```diff
@@ -124,6 +124,17 @@ func hostRunners(ctx context.Context, os string, redactions []*redact.Redact, l
...
 	r = append(r, lsMod)
 
+	//Dmesg runner
+	dMesg, err := host.NewDMesg(host.DMesgConfig{
+		OS:         os,
+		Timeout:    TimeoutTenSeconds,
+		Redactions: redactions,
+	})
+	if err != nil {
+		l.Error("unable to create host.Dmesg runner.", "err=", err)
```
groovemonkey, line 134: This is not something we need to fix in this PR, but I just noticed that the code for all the runner-creation here just ignores failures and adds failed runners to the runner list after logging. @nwchandler and @Ayeyiiat should I create a ticket to wrap all of the actual runner-slice appends in an `else` clause? E.g.
```
lsMod, err := host.NewLsmod(host.LsmodConfig{
	OS:         os,
	Timeout:    TimeoutTenSeconds,
	Redactions: redactions,
})
if err != nil {
	l.Error("unable to create host.Lsmod runner.", "err=", err)
} else {
	r = append(r, lsMod)
}
```

groovemonkey (approved):

product/host.go
nwchandler: Good catch, @groovemonkey! Yeah, if you could create a ticket for that, I think it would be worthwhile. Failures in any of these, right now, look like they'd probably cause a nil pointer panic.

nwchandler (approved):

PR #339, opened 2023-07-18
bugfix: prevent nil host runners from making it into the runner slice

groovemonkey:
This PR fixes the bug outlined in https://hashicorp.atlassian.net/browse/CORI2-1024

@Ayeyiiat and I chose to use the pattern of returning early that all other hcdiag Products use, instead of adding an `else` clause. Principle of least surprise; &c.


changelog/339.txt
```diff
@@ -0,0 +1,4 @@
+```release-note:breaking
```
nwchandler, line 1: Instead of `breaking`, I think this would just be `bug`. That'll group it into the right place in our changelog whenever that gets generated. And, then we can probably get rid of the bug prefix on line 2.

nwchandler (approved):
Good find and clean fix! I left a minor comment about the changelog, but the code looks good to me!

Ayeyiiat (approved):

changelog/339.txt
groovemonkey: I...am so sorry about how irresponsible I am when copying and repurposing changelog files. 😢 plz forgive

nwchandler: I forgive you! 😆

PR #350, opened 2023-12-19
Remove Jira Sync Workflow

nwchandler:
The Jira sync workflow is no longer needed, but it had been failing for some time, which caused CI to report errors. This merge removes the workflow to reduce this noise.


groovemonkey (approved):
I approve removal of all failing tests.

PR #351, opened 2024-02-16
upgrade to go 1.21.7

finnigja:
Moving to newest Go minor release, similar to previous https://github.com/hashicorp/hcdiag/pull/345.

There are a number of CVEs associated with 1.21.3 which hcdiag is not exposed to, but upgrading to avoid vulnerability scan noise.


kedron-abbott (approved):
LGTM. Will merge this and start the release process.

PR #356, opened 2024-07-30
Update Version to 0.5.4-rc0

kedron-abbott:
This PR adds a go version upgrade, and bumps the hcdiag version to 0.5.4-rc0 to push to staging

Jira Ticket: https://hashicorp.atlassian.net/jira/software/c/projects/CORI2/boards/548?assignee=712020%3A2169ba7a-45cb-47d5-9cfd-a660b6d01a06&selectedIssue=CORI2-1543


version/version.go
```diff
@@ -14,7 +14,7 @@ const (
 var (
 	// version is the main version number that is being run at the moment.
 	// version must be of the format <MAJOR>.<MINOR>.<PATCH>, as described in the semantic versioning specification.
-	version = "0.5.3"
+	version = "0.5.4"
```
nwchandler, line 17: This looks all right, but it's odd that GitHub shows no diff in the prerelease. The official 0.5.3 release ought to have had that removed, and I think it did. 🤔

nwchandler (approved):
Looks good, although there's something weird in the GitHub diff, I think. We need to be sure to get change the prerelease to "" before the final release.

version/version.go
kedron-abbott: I thought we did based on: https://github.com/hashicorp/hcdiag/pull/354/files

nwchandler: Yeah, I wonder why GH isn't showing that in the diff. I suppose it doesn't really matter for now, though.

kedron-abbott: For the Jira Sync I'm not sure how to resolve that error ( I tried adding the link to the description above)

Additionally, the docs say "The merge strategy you use SHOULD be “Create a Merge Commit,” so that we don’t accidentally lose the git history on this branch. It must not be a squash commit."

But it's not allowing me to.
image

Should I rebase and merge instead?

nwchandler: We should probably remove the Jira sync workflow from this branch. It was removed in `main` but never backported. That workflow is no longer needed. If you wanted to pull that change in (or just remove the workflow file), that PR is [here](https://github.com/hashicorp/hcdiag/pull/350).

I think that document probably needs updated. I don't recall when merge commits were disabled on the branch, but I think a rebase and merge would accomplish the same thing here. The main thing is not to squash PR merges in the branch.

kedron-abbott: Thank you!!

PR #361, opened 2024-09-27
Release v0.5.5

nwchandler:
This merge is the v0.5.5 release, whose only change from v0.5.4 is an update to Go 1.22 from Go 1.21.

Additionally, GitHub Actions have been updated as some were out of date and also not pinned to specific commits.


Ayeyiiat (approved):
LGTM!

PR #366, opened 2025-03-13
Add s390x support

bnferguson:
This PR is adding `s390x` support for `hcdiag` in support of some experiments and future projects where other tooling supports it but `hcdiag` does not yet. Since those projects would rely on this I figured it'd be good to get it started here. (was half tempted to add Power as well but held off :D) 

Also needed to update the deprecated `upload-artifact` GitHub Action to get things working.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
           version: ${{ needs.get-product-version.outputs.product-version }}
           product: ${{ env.PKG_NAME }}
           repositoryOwner: "hashicorp"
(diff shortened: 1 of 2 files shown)


nwchandler:
Hi, @bnferguson! Apologies for the delay. The hcdiag project is in maintenance mode at this time, and given what it does (collecting OS information, in addition to product info), I suspect it would not work as expected without some additional changes. I spoke with @cotarg about this as well, and we're going to close the PR. If this is a blocker for other work, please reach out via internal communications so we can come up with options together.

PR #369, opened 2025-04-01
run go mod tidy with 1.23.7

kedron-abbott:
Had to create this PR because the GH rules wouldn't allow me to push anything else to 0.5.7 nor would it allow me to delete it to redo it.


nwchandler (approved):
👍

